PR #2, opened 2014-09-14
Dlahqr2

### LINKED ISSUE #3: dneupd selects/calculates wrong eigenpairs if rvec = true
This is a followup discussion to the following bug: http://forge.scilab.org/index.php/p/arpack-ng/issues/1315/. I'm creating an issue here so that I can reference it from the PR on github.

thrasibule:
As discussed at length in http://forge.scilab.org/index.php/p/arpack-ng/issues/1315/, dneupd is broken when linked with dlahqr that comes from lapack-3.
Until someone knows how to port the code to the new lapack-3 functions, I brought back {d,s}lahqr from lapack-2 into the tree, and renamed them dlahqr2 and slahqr2 so that there is no symbol clash when linking with lapack-3.
I believe this should fix most of the issues observed in 1315 at the above url.
Closes #3

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,9 @@ arpack-ng - 3.2.0
  * Switch to github - https://github.com/opencollab/arpack-ng/
 
  -- Sylvestre Ledru <sylvestre@debian.org>  Sat, 17 May 2014 16:25:36 +0200
+ 
+ * Fix dsneupd select/calculate wrong eigenpairs if rvec = true
+   by using dlahqr and slahqr from lapack2 (Closes: #3)
 
 arpack-ng - 3.1.5
 


caliarim:
Hi all. I think I found the way to port the code to lapack 3+. Subroutines *laqrb.f are called by *neigh.f and they are "mostly a modification of the LAPACK routine *lahqr". I compared *laqrb.f with *lahqr.f (2.0) and found that the difference is that z is a 1d array set to zero-one at the beginning and later updated. You can get the same with

      do 5 j = 1, n-1
         bounds(j) = zero
  5   continue 
      bounds(n) = one

      call dlahqr(.true., .true., n, 1, n, workl, n, ritzr, ritzi, 1, 1, bounds, 1, ierr)

instead of 

 call dlaqrb (.true., n, 1, n, workl, n, ritzr, ritzi, bounds, ierr)

in dneigh.f (the same in sneigh.f, complex subroutines are safe). HTH,

Marco

thrasibule:
dlaqrb is not the issue here since it's provided in the source. We want a way to get the lapack 2.0 behaviour for dlahqr. Maybe we can emulate that by calling dlaqrb, but not the other way round.

caliarim:
The issue is "dneupd selects/calculates wrong eigenpairs if rvec = true". This is because the Ritz values computed by dnaupd are different from the eigenvalues computed by dneupd. And this is because the Ritz values computed by dnaupd are computed by dlaqrb, which is basically dlahqr 2.0, while the eigenvalues computed by dneupd are computed by the standard LAPACK dlahqr. Either you let dnaupd to use standard dlahqr (my way) or you let dneupd to use dlahqr 2.0 (current ARPACK behavior). If we believe that dlahqr 3.5.0 is an improved version of dlahqr 2.0, we should take the first way, unless someone has explicit counterexamples.

thrasibule:
I understand what you mean now, it makes sense. Let me run some tests and see if it fixes it, thanks.

Issue #14, opened 2015-06-22
dnsimp test fails in VPATH builds


rwirth:
Referencing http://forge.scilab.org/index.php/p/arpack-ng/issues/1430/
The dnsimp test fails when the build tree is not the source tree because a matrix file needed for the test is not present.

PR #15, opened 2015-06-22
execute the dnsimp test in the right directory

rwirth:
Fixes opencollab/arpack-ng#14


sylvestre:
This breaks the test. Could you fix the issue?

rwirth:
Forgot to add the script to `EXTRA_DIST`. Works now.

turboencabulator:
The AC_CONFIG_LINKS macro in configure.ac should already handle this, by making either a symlink or a copy of testA.mtx in the build directory.  This change should be unnecessary if that macro is working correctly.
This may have been a problem from before I added the macro.  Have you tried building from master lately?

sylvestre:
FYI, I simplified the patch in commit 82de5c6bed8a2600d311f1ca27dd843e12b5eacd

Issue #18, opened 2015-07-16
Parpack segmentation fault in pdsaupd -> pdlamch -> dgamx2d


davydden:
i experience a segmentation fault error in Parpack `3.1.4` and `3.2.0` compiled 
with `Scalapack 2.0.2` and `veclibfort`. I don't know whether this is an issue of 
Parpack, Scalapack or a mistake in a driving program. 

I get the SEGV error when using Parpack inside a FEM code deal.II https://github.com/dealii/dealii/pull/1146
but a more straight forward way to reproduce the problem is:
```
1. download a simple program from here (not written by myself)
http://forge.scilab.org/index.php/p/arpack-ng/issues/1480/
2. compile and run it even with 1 processor.
```

Expected result: pass

Actual result: 

```
Process 12397 stopped
* thread #1: tid = 0xb446d, 0x0000000115765ccc 
libscalapack.dylib`dgamx2d_ + 64, queue = 'com.apple.main-thread', 
stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000115765ccc libscalapack.dylib`dgamx2d_ + 64
libscalapack.dylib`dgamx2d_:
->  0x115765ccc <+64>: movq   (%rax,%rdi,8), %r14
    0x115765cd0 <+68>: movb   (%rdx), %bl
    0x115765cd2 <+70>: movb   %bl, %al
    0x115765cd4 <+72>: addb   $-0x41, %al
(lldb) bt
* thread #1: tid = 0xb446d, 0x0000000115765ccc 
libscalapack.dylib`dgamx2d_ + 64, queue = 'com.apple.main-thread', 
stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000115765ccc libscalapack.dylib`dgamx2d_ + 64
    frame #1: 0x00000001158f9157 libscalapack.dylib`pdlamch_ + 167
    frame #2: 0x000000011601d75b libparpack.2.dylib`pdsaupd_ + 2267
    frame #3: 0x0000000100027aac dvr_parpack`pdsaupd(n=256, nev=10, 
Evals=0x000000011f311fa0, Evecs=0x0000000120803200) + 1100 at 
dvr_parpack.cc:308
    frame #4: 0x0000000100026621 dvr_parpack`main(nr_arguments=2, 
arguments=0x00007fff5fbffa60) + 449 at dvr_parpack.cc:142
[...]

PR #19, opened 2015-08-25
Dlaqrb

thrasibule:
Issue #3 was partially solved with pull request #2, by bringing back an older version of dlahqr.
However this only worked with reference blas, but with atlas or openblas, the issue persists as evidenced by #7, #16 and #17. This is another attempt at fixing it using a strategy suggested [here](http://forge.scilab.org/index.php/p/arpack-ng/issues/1315/#ic4079) by Marco Caliari.

This is an attempt to close #7 ,close #16  and close #17.

PR #21, opened 2015-08-30
rename pdlamch to pdlamch10 to avoid symbol collision with Scalapack …

davydden:
…2.0.2 where this function is different.

should fix https://github.com/opencollab/arpack-ng/issues/18 but until https://github.com/opencollab/arpack-ng/pull/19 is merged I can't test it on my mac.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,6 @@
 arpack-ng - 3.3.0
+ [ Denis Davydov ]
+ * Rename pdlamch to pdlamch10 to avoid symbol collision with Scalapack 2.0.2 in MPI context.
 
  [ Kyle Guinn ]
  * General improvements on the build system
--- a/PARPACK/SRC/MPI/Makefile.am
+++ b/PARPACK/SRC/MPI/Makefile.am
@@ -20,7 +20,7 @@ EXTRA_DIST = debug.h stat.h
 
 noinst_LTLIBRARIES = libparpack_noopt.la libparpacksrcmpi.la
 
-libparpack_noopt_la_SOURCES = pslamch.f pdlamch.f
+libparpack_noopt_la_SOURCES = pslamch.f pdlamch10.f
 libparpack_noopt_la_FFLAGS = -O0
 
 libparpacksrcmpi_la_SOURCES = $(SSRC) $(DSRC) $(CSRC) $(ZSRC)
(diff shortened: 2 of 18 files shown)


davydden:
sorry to bother, could you give a quick look at this PR as it should be an easy fix of a serious bug @sylvestre .

sylvestre:
Sorry, I missed your initial PR.
A few questions:
* Why arpack-ng has to be renamed? Why not scalapack? Looks at Debian statistics, arpack (even parallel arpack) has much more users than scalapack 
* What will happen to software already using pdlamch?
* Why pdlamch10? (the 10 ?)
* The soname dump is missing as the ABI changed.

For now, I don't think we can rename this function. The impact might too big.

davydden:
*  i think `arpack-ng` has to be renamed as it is actually a copy-pasted function (veresion `1.0`) from `scalapack` to be used inside `arpack` software. To my understanding it was not written by arpack authors. During last 10-15 years this function changed upstream as scalapack evolved. So i think it makes sense to renamed the copy-pasted function to something else.

*  if there is a software which used `pdlamch` as provided by `arpack` then it will certainly be broken. But i would argue that it's a bad practice anyway. I think the issue here is that `arpack` should not have copy-paste the function but declare a dependency (requirement) of `scalapack` for build and then use inside this single function. Then this issue would have been obvious the moment `scalapack` interface changed. 

* `pdlamch10` because it's version `1.0`. Could as well be 99, but i thought `10` makes more sense.

Keep in mind that if you don't rename this function then the whole arpack is somewhat broken as it would lead to segmentation fault in certain scenariou (which is very hard to trace!) as i discussed in the issue https://github.com/opencollab/arpack-ng/issues/18 . 
Possible user cases is combining `arpack` and `mumps` (a popular parallel sparse direct solver) or any other solver for that matter which is build against `scalapack`. I would be happy to rework the PR if you have any other solution in mind to avoid symbol conflict.

sylvestre:
No, your rational is fine! Thanks for taking the time to explain.

sylvestre:
I guess it is too hard to write a non regression test for this?

davydden:
i am not sure you can do any tests here because the problem appears not when you compile `arpack` (at that stage even all tests could be fine), but when you actually link some executable which happen to have `libarpack` and `libscalapack` in its linking flags.

So if the user (or a third party library) is clever to check for symbol clash (and if so disable arpack or another library), then the segmentation fault can be avoided. But I don't think you can do it on `arpack` side.

I don't know Fortran much, perhaps there could be some low-level solutions how to avoid the issue, but supposedly they would be equivalent to a simple rename.

sylvestre:
Ok, thanks!

PR #26, opened 2015-10-14
cmake support

m-reuter:
I included a CMakeLists.txt file (adopted from https://github.com/feymark/arpack.git) to support cmake out of code builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
 
 script:
   # CMAKE
-  - if [ "$BUILD" == "cmake"  ]; then make ; fi
+  - if [ "$BUILD" == "cmake"  ]; then make all check ; fi
   # CONFIGURE
   - if [ "$BUILD" == "configure" ]; then ./bootstrap ; fi
   - if [ "$BUILD" == "configure" ]; then ./configure --enable-mpi ; fi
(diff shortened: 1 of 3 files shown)


sylvestre:
Do you have the agreement from @feymark to reuse his work ? 
I don't see a license in the cmake file.

Could you mind updating travis to also run the build with cmake? Thanks

feymark:
You could use it as you want. No problem by me.

sylvestre:
Great, thanks.

By the way, the docs should be updated. For now, I guess we can live with the two build systems in parallel.

m-reuter:
Added cmake build to travis. The cmake build compiles, the old configure build seems to be failing (although I changed nothing about that). Also the cmake does currently build static only. Furthermore, it does not yet include the checks.

m-reuter:
Hi, I updated the Readme to document the parallel cmake support, and also added the checks to the makefile and to travis. So travis now compiles library, examples, parallel (MPI) and runs checks.
The cmake build compiles and runs fine and much faster than the configure build (which gets created in travis in parallel as a second thread). From my side I think this is fine now and could be merged. Also note that the cmake part never fails travis, even with the openblas (at least not yet for me), while the configure build failed sporadically with openblas.

sylvestre:
This is really great, thanks! I see a few missing things (like the cmake file not shipped in the make dist or the changelog not updated) but no big deal, we can fix that later. Thanks again!

PR #31, opened 2016-01-08
Autotools fixes

turboencabulator:
I have another batch of autoconf/automake fixes.  The most important change in here is fixing the order of BLAS/LAPACK link flags.  The old order (-lblas -llapack) won't work in some situations, particularly if they are static libs.  The other commits are mostly cleanup and simplification.


sylvestre:
BTW, I am considering to remove the autotools files and move to cmake ( https://github.com/opencollab/arpack-ng/blob/master/CMakeLists.txt ) , what do you think?

turboencabulator:
Let me start out by saying I'm a Unix guy, and cmake has always seemed like a foreign way to do things on Unix -- it doesn't seem to integrate that well.  I think it caters more toward the Windows and Mac communities, and it focuses too much on making the output pretty (colorized output, progress indicators) rather than useful (showing the actual compiler commands).

I just tried it out.  Here are a few examples of how it doesn't integrate very well:
* Only builds a static lib.
* I found the magic flag to build a shared lib (`-DBUILD_SHARED_LIBS=ON`).  Unlike with autotools, you can't build both static and shared at the same time.
* The shared lib doesn't have the same SONAME as with autotools.
* Doesn't create arpack.pc.
* `make install` doesn't work.

But I do see the other issues where some people can't or don't want to set up autotools.  I have a feeling that those people are in the Windows and Mac communities and would prefer cmake.  If you don't have a good way to do a `make dist` and host the tarball somewhere, then having cmake support instead may keep those users happy.

So I'd prefer keeping both to satisfy everyone.  I can continue to maintain autotools.  It shouldn't be much effort since the project is in maintenance mode.

nalimilan:
If you give a look to the Wikipedia page on CMake, you'll see that many projects which are not Windows- or Mac-oriented use it (including LLVM, Qt, LAPACK, libpng, cURL...). I think the main motivation behind this is to simplify the build system, which is often considered obscure with the autotools. Build performance is also said to be better with CMake.

tkelman:
CMake is a huge step above lapack-style "fill in this makefile with your system info." It can be a bit more effort to get up to standard autotools parity though - it's capable of doing both shared and static libraries, or adding SONAME, creating pkg-config files, setting up `make install`, etc but you have to add the right steps to the CMakeLists to do each of those. For a project already invested in autotools the big benefit CMake brings is support for Visual Studio, but since Arpack is in Fortran that's not so relevant here. It would likely support the Windows version of the Intel Fortran compiler better than autotools does, if that's something people care about here.

turboencabulator:
I hate to throw out something that works well (for me, at least) and replace it with something that isn't as polished.  It's not like the cmake and autotools files can't coexist and we have to choose only one.  We still have Visual Studio project files in the repo and they don't appear to be causing any trouble.  If your preferred build system isn't up to parity with the others, send a pull request.  That's what I've been doing :-)

Issue #30, opened 2016-01-06
Allow calling BLAS functions with suffix to use ILP64 BLAS


nalimilan:
Some programs like Julia and Numpy have started to (or tried to) use ILP64 BLAS in order to get rid of size limitations on arrays. But this creates conflicts with libraries that expect a LP64 BLAS when they are loaded in the same process (see https://github.com/JuliaLang/julia/issues/4923, https://github.com/numpy/numpy/issues/5906 and linked issues). The solution that has been found is to export a different BLAS API, by adding a suffix to all functions and to the library name. OpenBLAS now supports a build flag for that (https://github.com/xianyi/OpenBLAS/pull/459), which is used in Fedora to ship an ILP64 package.

The most common suffix is `64_`, as it's already used by several programs (https://github.com/xianyi/OpenBLAS/issues/646). SuiteSparse offers the `-DSUN64` build flag to use the suffixed API.

Would you consider supporting a similar API? It appears to be possible with the current build system. Julia uses the following Makefile instructions: https://github.com/JuliaLang/julia/blob/69fb335efe6975c52f8b3d1c66d4008ce2d0b1cb/deps/Makefile#L1316-L1332

It would be great if these could be included upstream so that a single flag needs to be switched. That would make it easier for packagers to ship both the LP64 and ILP64 versions (https://bugzilla.redhat.com/show_bug.cgi?id=1294201).

sylvestre:
Sure. Happy to simplify your life and this sounds an interesting change.

Could you propose a pull request implementing this change?
Thanks

nalimilan:
Cool. I'll see what I can propose. But I have no idea how to do this in CMake too. Is that a problem?

sylvestre:
Well, the sad thing is that we have both autotools and cmake for now :) I would like to kill autotools at some points.
Anyway, I don't think it is hard to do in CMake.

nalimilan:
I've made a PR with changes to the autotools code. Please tell me how you feel about it, and if it's OK I'll have a look at how to do this with CMake.

FWIW, a typical `configure` call would look like this on Fedora:
```
./configure LIBSUFFIX=64_ SYMBOLSUFFIX=_64 INTERFACE64=1 \
--with-blas=openblasp64_ --with-lapack=openblasp64_
```

[...]

PR #32, opened 2016-01-16
ILP64 support

nalimilan:
Cf. https://github.com/opencollab/arpack-ng/issues/30

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+arpack-ng - 3.4.1
+  [ Milan Bouchet-Valat]
+  * Allow adding suffixes to symbols and library names to build ILP64 version
+    based on ILP64 BLAS/LAPACK with suffixes. This avoids conflicts when loading
+    libraries with different integer sizes in the same program.
+
 arpack-ng - 3.4.0
 
  [ Martin Reuter ]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.6)
 project(arpack)
 
 # Adopted from https://github.com/feymark/arpack.git
@@ -9,6 +9,11 @@ option(MPI "Enable parallel support" OFF)
 #option(SYSTEM_BLAS "Use system BLAS" ON)
 #option(SYSTEM_LAPACK "Use system LAPACK" ON)
 option(EXAMPLES "Compile ARPACK examples" OFF)
+set(LIBSUFFIX ""
+    CACHE STRING "suffix to add to ARPACK libraries names")
+set(SYMBOLSUFFIX ""
+    CACHE STRING "suffix to add to ARPACK, BLAS and LAPACK function names")
+option(INTERFACE64 "use the 64-bit integer interface (ILP64) for ARPACK, BLAS and LAPACK")
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
@@ -104,7 +109,7 @@ add_library(arpack
 
 target_link_libraries(arpack ${BLAS_LIBRARIES})
 target_link_libraries(arpack ${LAPACK_LIBRARIES})
-set_target_properties(arpack PROPERTIES OUTPUT_NAME arpack)
+set_target_properties(arpack PROPERTIES OUTPUT_NAME arpack${LIBSUFFIX})
 
 if (MPI)
 #    add_library(parpack SHARED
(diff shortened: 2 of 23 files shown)


sylvestre:
The build is failing with this change, could you have a look?
By the way, are you also planning to update the cmake build? Thanks

nalimilan:
The build failure looks like I missed a substitution in the Makefiles somewhere, but I need to find where. I hadn't tested PARPACK correctly.

I'll take care of CMake if you agree with the general strategy.

sylvestre:
Looks great. Could you just add some automation and testing about this in travis?
https://github.com/opencollab/arpack-ng/blob/master/.travis.yml

Maybe a new build with ILP64 and the tests.

nalimilan:
It would be nice, but we need an ILP64 BLAS with suffix to test this, and I don't think Debian/Ubuntu has packages for that...

sylvestre:
too bad. A solution would be to build also blas in travis but this might be overkill

nalimilan:
OK, I found the bug, and now it looks good.

That said, the tests fail with ILP64 (this also happens in the Julia tree). Correct output with LP64:
```
$ TESTS/bug_1315_double 
1000.000000
999.000000
998.000000
997.000000
996.000000
995.000000
994.000000
993.000000
992.000000
```
Incorrect output with ILP64:
```
$ deps/build/arpack-ng-3.2.0/TESTS/bug_1315_double 
1000.000000
999.000000
998.000000
997.000000
995.000000
```
I suspect this is because of the tolerance. Could you suggest a fix? Could be left for another PR though, as it isn't strictly related to the present set of changes.

Finally, I've left a FIXME regarding the way to set the Fortan integer size. How could I adapt the argument depending on the compiler?

nalimilan:
Actually, I've found a way to increase portability (see the last commit).

sylvestre:
I merged another PR which conflict with yours, could you rebase it? Sorry about that :/

sylvestre:
Btw, are you planning to update the cmake build here?

sylvestre:
Also, please update the changelog with your changes

nalimilan:
Will do. I'm currently looking at how to get this to work with CMake. One of the issues is that I haven't found a way to apply the Fortran name mangling from the CMake scripts: the only supported way seems to be to create a C header file. This in turns means that 1) the header must be created even when not adding a suffix, and 2) that I need to adapt the Makefiles too.

nalimilan:
I've pushed another commit with CMake support. Everything looks fine, but I couldn't fully test it because AFAICT I can't specify a custom BLAS.

There's also the tests failure mentioned at https://github.com/opencollab/arpack-ng/pull/32#issuecomment-172224682. Any pointers?

@tkelman @stevengj Can you confirm that this would be OK for Julia? In particular, `OPENBLAS_SYMBOLSUFFIX=64_` will have to be translated to `SYMBOLSUFFIX=_64` here, but I don't think we can do things differently if we want to be robust to name mangling differences between compilers.

tkelman:
If you've tested it, I'll take your word for it. I'd rather spend effort on replacing arpack than patching it, but that's just me.

stevengj:
@nalimilan, that sounds right.

nalimilan:
@stevengj Any idea why the tests for bug 1315 are failing with ILP64? I saw you wrote the patch to use `BLASINT`, but that doesn't sound enough to make them work. IIUC, the tests are supposed to return 1000, 999, 998..., but they return 733.66, 0, 0... I don't see what could be wrong, except a wrong integer size somewhere (but where?).

nalimilan:
@sylvestre Do you have any ideas about how to fix this? Other than that, the PR should be ready.

sylvestre:
Sure, could you open a new bug about the CI integration? So that we don't forget when Debian/Ubuntu proposes this.

nalimilan:
@sylvestre Thanks, but the tests are still failing with ILP64 enabled. Could you give me any pointers?

I've filed an issues about TravisCI at https://github.com/opencollab/arpack-ng/issues/33, but there's no guaranty that Ubuntu will include an ILP64 OpenBLAS.

BTW, any idea when you'll publish a new release?

ViralBShah:
Apologies for chiming into such an old PR. 

I notice that `znaupd`, `zneupd`, `cnaupd`, and `cneupd` are included for name mangling with SYMBOLSUFFIX in the CMakeLists.txt only for Fortran and not for C. That means that the complex valued functions do not end up with the suffixed versions in the generated library.

Example build script for Julia's bundling of ARPACK: https://github.com/JuliaPackaging/Yggdrasil/blob/4a60c14c81f8f5752b19b23cb6b4e040a1b6ecc2/A/Arpack/build_tarballs.jl
```
➜  arpack-ng git:(master) nm /Users/viral/.julia/artifacts/87b6fe9f756585c40567136a5149f17b72856942/lib/libarpack.2.1.0.dylib| grep -i cnaupd
0000000000019570 T _cnaupd_
➜  arpack-ng git:(master) nm /Users/viral/.julia/artifacts/87b6fe9f756585c40567136a5149f17b72856942/lib/libarpack.2.1.0.dylib| grep -i dnaupd
000000000001ffc0 T _dnaupd_64_
```

nalimilan:
At this point I have no idea whether this was just an oversight. Maybe make a PR to add them to the list?

ViralBShah:
The `SCALARFUNS` and `COMPLEXFUNS` have the mangling defs added to `CMAKE_Fortran_FLAGS` whereas `CFUNS` are added to `CMAKE_C_FLAGS`.

fghoussen:
> The SCALARFUNS and COMPLEXFUNS have the mangling defs

Use ICB: you won't need that anymore. Before ICB, symbols where defined "by hand" which is not portable.
https://github.com/opencollab/arpack-ng/tree/master#iso_c_binding-support

ViralBShah:
We basically need BLAS and LAPACK calls with `_64` suffixes. I don't think ICB will help with that, IIUC.

fghoussen:
> I notice that znaupd, zneupd, cnaupd, and cneupd ...

Only arpack `*n[ae]upd` symbols are concerned by the arpack-ng ICB option.

> We basically need BLAS and LAPACK calls with `_64` suffixes. I don't think ICB will help with that, IIUC.

BLAS, LAPACK can provide, or not, portable (ICB) or non-portable symbols (ideally each library should. For BLAS, LAPACK this is a huge and error-prone work with high impact: not sure if they provide ICB symbols, but, considering the involved work/impact, I'd understand they don't).

fghoussen:
> The SCALARFUNS and COMPLEXFUNS have the mangling defs added

Ideally the "old way" of exporting (non-portable/exhaustive) symbols (using `SCALAR/COMPLEXFUNS) should be suppressed. ICB is the "new" portable (according to latest fortran standard) way to export symbols.

fghoussen:
> I notice that znaupd, zneupd, cnaupd, and cneupd are included for name mangling with SYMBOLSUFFIX

Likely because all these symbols have not been "added by hand" here:

```
>> git grep COMPLEXFUNS
CMakeLists.txt:        set(COMPLEXFUNS, naitr napps naup2 naupd neigh neupd ngets statn getv0 sortc)
configure.ac:    COMPLEXFUNS="naitr napps naup2 naupd neigh neupd ngets statn getv0 sortc"

```

Try to add them here. If this creates expected symbols, you may even PR this here.

Issue #34, opened 2016-02-23
pdlamch10 return variable not correct after renaming pdlamch


wstrydom:
The function pdlamch10 ends with the following code:
```
      PDLAMCH = TEMP
*
*     End of PDLAMCH
*
      END
```
Since the variable was not renamed to PDLAMCH10, the return variable is not set and the correct tolerance values are not being used.

PR #35, opened 2016-03-14
fix pdlamch10 return variable

davydden:
fixes https://github.com/opencollab/arpack-ng/issues/34

in reference to this PR https://github.com/opencollab/arpack-ng/pull/21


sylvestre:
Could you write a test to make sure it doesn't come back? Thanks

davydden:
@sylvestre sorry, but I do not know Fortran and this function well enough to create a unit test.

Issue #38, opened 2016-07-06
[3.4.0] cmake build shortcomings


davydden:
as of now cmake build miss the following important features:

- `install` target with `CMAKE_INSTALL_PREFIX`

- a way to specify Blas/Lapack libraries

- building `shared` libraries is broken due to a missing dependency of parpack on arpack.

davydden:
i will look into adding an install target...

PR #39, opened 2016-07-06
Improve cmake build

davydden:
fixes https://github.com/opencollab/arpack-ng/issues/38

tested on macOS.

@m-reuter @nalimilan: you might want to give a quick look at these Cmake-related changes.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,7 @@
 arpack-ng - 3.5.0
 
+ [ Denis Davydov ]
+ * Improve cmake build system: add make install and fix shared libraries.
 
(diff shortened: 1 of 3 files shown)


sylvestre:
Many thanks. I am currently in holidays but it will review this when I get back

nalimilan:
I'm really not an expert as regards CMake, so i let you review this. Though I'd appreciate it if you could check that it still works with `INTERFACE64=1 SYMBOLSUFFIX=_64 LIBSUFFIX=_64` when building against an ILP64 suffixed BLAS (I don't see any reason why it would break, though).

sylvestre:
Thanks again!

Issue #41, opened 2016-07-25
3.4.0 Libtool wrongly set for cygwin


matzeri:
On Cygwin, latest 3.4.0 arpack sets libtool wrongly. 
Comparing 3.3.0 libtool version with 3.4.0 on the build trees, I have (as extract)

-libext=a
+libext=lib

-shlibpath_overrides_runpath=yes
+shlibpath_overrides_runpath=unknown

-library_names_spec="\$libname.dll.a"
+library_names_spec="\$libname\`echo \$release | \$SED -e s/[.]/-/g\`\$versuffix\$shared_ext \$libname.lib"

-soname_spec="\`echo \$libname | sed -e s/^lib/cyg/\`\`echo \$release | \$SED -e s/[.]/-/g\`\$versuffix\$shared_ext"
+soname_spec=""

It seems libtool is incorrectly set for MSVC platform
As the changes on configure.ac are mainly base on the addition of "ax_check_compile_flag.m4" 
and its call,  I guess the root cause is there but I was unable to identify the exact culprit.

$ libtool --version
libtool (GNU libtool) 2.4.6
$ automake --version
automake (GNU automake) 1.15

sylvestre:
Having a test patch would be appreciated, I don't have a way to test this.

matzeri:
It will take time to have a patch. I will need to bisect the configure.ac new tests to see which one is causing the fault. May be is a corner case of Autoconf/Automake. I never had before such problem on any package that I am packing for cygwin (~ 60 use Autoconf/Automake)

matzeri:
the problem is caused by a single line, removing it all return to normal
```
--- origsrc/arpack-ng-3.4.0/configure.ac        2016-07-02 21:52:15.000000000 +0200
+++ src/arpack-ng-3.4.0/configure.ac    2016-07-29 09:39:40.152241500 +0200
@@ -37,7 +37,6 @@ AC_ARG_VAR(SYMBOLSUFFIX, [suffix to add

 if test x"$SYMBOLSUFFIX" != x""; then
   dnl Need to rely on non-F77 features
-  AC_FC_LINE_LENGTH(unlimited)
   FFLAGS="$FFLAGS $ac_cv_fc_line_length"

   AX_CHECK_COMPILE_FLAG(-cpp, FFLAGS="$FFLAGS -cpp",
```

sylvestre:
Good, could you write a pull request to fix that? Thanks

matzeri:
[...]

PR #42, opened 2016-07-29
removing AC_FC_LINE_LENGTH to solve #41 on Cygwin

matzeri:
Tested on latest Cygwin 32 and 64bit. No test failure.

No idea why "AC_FC_LINE_LENGTH(unlimited)" was introduced but I guess mixing F77 with FC  commands is confusing Autoconf at least on Cygwin in this case

$ gfortran --version
GNU Fortran (GCC) 5.4.0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,7 +37,6 @@ AC_ARG_VAR(SYMBOLSUFFIX, [suffix to add to ARPACK, BLAS and LAPACK function name
 
 if test x"$SYMBOLSUFFIX" != x""; then
   dnl Need to rely on non-F77 features
-  AC_FC_LINE_LENGTH(unlimited)
   FFLAGS="$FFLAGS $ac_cv_fc_line_length"
 
   AX_CHECK_COMPILE_FLAG(-cpp, FFLAGS="$FFLAGS -cpp",
--- a/configure.ac
+++ b/configure.ac
@@ -7,7 +7,7 @@ AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_LINKS([TESTS/testA.mtx:TESTS/testA.mtx])
 
 dnl Checks for standard programs.
-AC_PROG_F77
+AC_PROG_FC
 AC_PROG_CC
 
 dnl Check for BLAS libraries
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -37,7 +37,6 @@ AC_ARG_VAR(SYMBOLSUFFIX, [suffix to add to ARPACK, BLAS and LAPACK function name
 
 if test x"$SYMBOLSUFFIX" != x""; then
   dnl Need to rely on non-F77 features
-  AC_FC_LINE_LENGTH(unlimited)
```
sylvestre: I think this is here for a reason. I don't remember why but we should probably just disable this line and the next when we detected we are on cygwin instead of removing it for everybody

tkelman:
It was introduced in https://github.com/opencollab/arpack-ng/pull/32/files#diff-67e997bcfdac55191033d57a16d1408aR40 since adding a symbol suffix via the preprocessor can result in some lines ending up longer than the default fixed-format length.

matzeri:
As it is a AC_FC,   should have included the replacement of
-AC_PROG_F77
+AC_PROG_FC

you can drop this pull, but replace AC_PROG_F77.

tkelman:
Yes, it does look like the mismatch was confusing autoconf on some platforms.

> you can drop this pull, but replace AC_PROG_F77.

What do you mean?

matzeri:
I found an alternative solution. Replacing
-AC_PROG_F77
+AC_PROG_FC
and leaving
AC_FC_LINE_LENGTH(unlimited)

also works on cygwin.
I guess AC_FC_LINE_LENGTH requires AC_PROG_FC ans it is not compatible with AC_PROG_F77

matzeri:
https://github.com/matzeri/arpack-ng/commit/a296e78f11e1bbd41e2a478dcf6fd2327876901c

tkelman:
since that's on the same branch as this PR was opened from, that second commit is now included here too

sylvestre:
I am concern it might have unexpected side effects on other platforms, what do you think?

sylvestre:
Probably not. @matzeri Could you update the CHANGES file to explain your change? Thanks

ValZapod:
> Is anyone still using F77-only compilers like g77 or f2c? Does autotools still work in those situations with FC?

igraph uses it.

PR #45, opened 2016-09-13
Disable C++ compiler detection

jschueller:
& set default build type to release


jschueller:
@sylvestre ok

sylvestre:
Oh, I forgot, could you detail the change in the CHANGES file? Thanks

Issue #46, opened 2016-09-13
p*napps cause deadlock when using custom communicators


iskakoff:
If the arnoldi runs on a custom communicator created for some subset of available cpus and after successful completion the second execution of the arnoldi with a communicator for a larger number subset of available cpus it will cause deadlock due to the following condition in code:

`data       first / .true. /`
`c`
`c     %-----------------------%`
`c     | Executable Statements |`
`c     %-----------------------%`
`c`
`      if (first) then`
`... `
`     first = .false.`
`...`
`endif`

the variable `first` is set to `false` in the first subset of cpus and will avoid a condition body evaluation, but other cpus have `first` equals to `true` and will evaluate condition body. Inside the condition body there is a `p*lamch10` call where the `MPI_ALLREDUCE` is called.

the possible workaround is to perform `MPI_ALLREDUCE(...MPI_LAND...)` for the variable `first` to be sure that all cpus inside communicator have the same value. But maybe there is a better solution.

sylvestre:
It would be nice if you could try a fix (or at least a test to help other volunteers to reproduce it).

sylvestre:
Many thanks!

PR #48, opened 2016-09-19
Fix a potential deadlock when using custom communicators (Fix issue #46)

iskakoff:
Possible solution for https://github.com/opencollab/arpack-ng/issues/46

remove variable `first` from `p*aitr` and `p*apps`, create two  `common` variables `apps_first` and `aitr_first`. Each time when `p*aupd` is called for a first time (`ido .eq. 0`) set variables `*_first` to `.true.` then they will be updated in `p*aitr` and `p*apps`.


sylvestre:
Could you add a test which demonstrates the issue you are addressing? (to make sure we don't regress) 
Thanks
(and also update CHANGES)

iskakoff:
Ok, will add test soon.

sylvestre:
Thanks!

sylvestre:
With an update of CHANGES too, we will be perfect. Thanks again!

iskakoff:
Done

PR #53, opened 2016-12-04
Improvements to CMake

### LINKED ISSUE #44: PkgConfig file?
I'm installing Arpack 3.4.0 using CMake on OSX but I'm missing the pkgconfig file `arpack.pc`. It's installed when I used `configure`, but not `cmake`. Any ideas?

Thanks.

turboencabulator:
Add some feature parity with autotools.
Attempt to fix #44 but this still needs some work before arpack.pc can be installed.


sylvestre:
Could you update the changelog ? thanks

Issue #58, opened 2017-03-29
NaN eigenvector in shift-invert mode


caliarim:
Hi, the attached file (a modification of dndrv2) tries to compute eigenvalues and eigenvectors of a slightly modified identity matrix (which turns out to be singular). Applying a proper shift makes OP non-singular. However, the eigenvector corresponding to the eigenvalue 0 is a column of NaN. Even if I do not know exactly what happens, I would propose a simple fix. At the end of dneupd.f, there is the purification stage (do 110) in which some divisions by real numbers take place. The problem is that those numbers can be zero, as in the attached example. My proposal is to avoid the purification if it is the case. I would rewrite the block as
```
 do 110 j=1, nconv
            if ((workl(iheigi+j-1) .eq. zero) .and.
     &           (workl(iheigr+j-1) .ne. zero)) then
               workev(j) =  workl(invsub+(j-1)*ldq+ncv-1) /
     &                      workl(iheigr+j-1)
            else if (iconj .eq. 0) then
               temp = dlapy2 ( workl(iheigr+j-1), workl(iheigi+j-1) )
               if (temp .ne. zero) then
                  workev(j) = ( workl(invsub+(j-1)*ldq+ncv-1) * 
     &                 workl(iheigr+j-1) +
     &                 workl(invsub+j*ldq+ncv-1) * 
     &                 workl(iheigi+j-1) ) / temp / temp
                  workev(j+1) = ( workl(invsub+j*ldq+ncv-1) * 
     &                 workl(iheigr+j-1) -
     &                 workl(invsub+(j-1)*ldq+ncv-1) * 
     &                 workl(iheigi+j-1) ) / temp / temp
               end if
               iconj = 1
            else
               iconj = 0
            end if
 110     continue
```
I think it is a safe fix, it cannot do worse than dividing by zero.
[dndrv2_47236.f.zip](https://github.com/opencollab/arpack-ng/files/878276/dndrv2_47236.f.zip)

sylvestre:
Please make a pull request with a test, thanks!

PR #59, opened 2017-03-29
Test for issue #58

caliarim:
I added a test is TESTS/bug_58_double.f


sylvestre:
The test needs to be added to https://github.com/opencollab/arpack-ng/blob/master/TESTS/Makefile.am
and to https://github.com/opencollab/arpack-ng/blob/master/CMakeLists.txt to be executed

sylvestre:
andhttps://github.com/opencollab/arpack-ng/blob/master/CHANGES should be also updated

caliarim:
OK, thanks for the explanation. I will do.

PR #66, opened 2017-08-19
fix {d|s}lahqr comment, which should be LAPACK routine

adamadanandy:
DLAHQR and SLAHQR should be LAPACK routine
The comment is fixed in dneigh.f and sneigh.f


sylvestre:
Thanks!

PR #67, opened 2017-08-30
correct spelling mistakes

EdwardBetts:


sylvestre:
great, what did you use to do that? thanks

EdwardBetts:
codespell: https://pypi.python.org/pypi/codespell/

PR #68, opened 2017-09-12
Add support for iso_c_binding (portability, f/c/c++ binding).

fghoussen:
I added support for iso_c_binding (helps portability - interfacing c/fortran since std=f2008): I supposed you could be interested into this. Even if you don't pull the code (as validation is KO on travis), you may want to do it your own way (changing file names, tests, ...) based on my commit as an example.

Regarding validation (make check) :

1. my laptop (debian): OK with autotools (with --enable-mpi and/or --enable-icb), but, KO with cmake.
2. cmake validation: my understanding (?) is that this didn't work (both laptop/travis) even before I cloned the repo.
3. travis validation: KO both with cmake/autotools... I can't reproduce the problem (when using --enable-mpi with autotools) on my laptop so can't fix it...

I have commited and pulled to you to let you know: you can pull or trash it !...

Hope this can help.

Franck


CMakeLists.txt
```diff
@@ -421,6 +421,7 @@ if(ICB)
   add_test(iso_c_binding_c Tests/iso_c_binding_c)
   set(CHECK_EXE "${CHECK_EXE} iso_c_binding_c")
 
+  enable_language(CXX)
```
sylvestre: I don't think we want that to be unconditionally set.
I guess most of the users of arpack won't enable this option, we should not force them to need a C++ compiler then.

fghoussen:
This will be set only if asked by the user (--enable-icb for autotools, or, -DICB=ON for cmake). By default, this is off. If on, one need a c++ compiler for testing iso_c_binding for c++. This is done in the same spirit than the *_f provided in hdf5 for example.

CMakeLists.txt
fghoussen: By the way, this is not unconditionally set: this is conditioned by a "if(ICB)". ICB must be turned on at cmake time. if on, one need a c++ compiler for testing iso_c_binding for c++ (I added 2 tests: one for fortran/c binding, and, another one for fortran/c++ binding)

fghoussen:
I got the configure validation OK on travis (make check OK after configure with --enable-mpi or --enable-icb): see 1b8cf4c

sylvestre:
To avoid any incorrect expectation, cmake was working as you can see in the last merged PR:
https://travis-ci.org/opencollab/arpack-ng/builds/270146014

This is why I didn't review that patch.

fghoussen:
I rebased the previous 3 commits in 1 for clarity. I fixed the CMakeLists.txt: autotools and cmake validations are both OK (see travis attached to 5b69cca).

If you are not convinced by benefits you could get from iso_c_binding, have a look at this http://www.idris.fr/media/formations/fortran/idris_fortran_2003_cours.pdf (section 9). iso_c_binding is meant to improve portability.

Franck

sylvestre:
Thanks. I am going to merge it. Could you just please mention the change in the CHANGELOG?
I would appreciate some examples too.

fghoussen:
I have added a few lines in the CHANGES file.
2 examples are provided (and have been added to autotools and cmake test suite):
- TESTS/iso_c_binding_c.c (C/Fortran binding using ssaupd)
- TESTS/iso_c_binding_cpp.cpp (C++/Fortran binding dsaupd)

sylvestre:
Sorry but tests are not examples ...
For example, we have this currently:
https://github.com/opencollab/arpack-ng/tree/master/EXAMPLES/SYM

fghoussen:
The only example I am working on uses PETSc (CSR matrix) for matmult operations: can't add PETSc as prerequisite for arpack-ng ! Writing an example with dense matrices (LAPACK) will be as tricky (no doubt) as it was to get PETSc work (I won't be able to justify that ! This is off-work contribution)... Adding as example another flavour (csaupd) of the 2 tests could be a trade-off : is this OK with you ?

sylvestre:
I was just asking for a simple trivial example, nothing complex :)
The tests with more comments and explanations would be enough. thanks!

fghoussen:
Is this OK with you ? (README and comments added) Franck

sylvestre:
Perfect, thanks

PR #69, opened 2017-10-14
Provide find_package for (cmake) users to find arpack-ng.

fghoussen:
This is the cmake equivalent of pkg-config.

The make install will ship 2 config files:
/path/to/arpack-ng/BUILD> cmake ..; make install
[100%] Built target arpack
Install the project...
-- Install configuration: "Release"
-- Installing: /path/to/arpack-ng/local/lib/libarpack.a
-- Installing: /path/to/arpack-ng/local/ArpackNGConfig.cmake
-- Installing: /path/to/arpack-ng/local/ArpackNGConfigVersion.cmake

Now, the user who wants to use arpack-ng can use find_package:

/tmp> export CMAKE_PREFIX_PATH="/path/to/arpack-ng/local:${CMAKE_PREFIX_PATH}" # Should be done by module load arpack-ng.

/tmp> more CMakeLists.txt
cmake_minimum_required(VERSION 3.7)
find_package(ArpackNG 3.5 REQUIRED) # Find *Config*.cmake.

/tmp> mkdir BUILD; cd BUILD; cmake ..
...
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/BUILD

/tmp/BUILD> cmake -LA .. | grep Arpack
ArpackNG_DIR:PATH=/path/to/arpack-ng/local
ArpackNG_INCLUDE_DIRS:PATH=/path/to/arpack-ng/local/include
ArpackNG_LIBRARIES:FILEPATH=/path/to/arpack-ng/local/lib/libarpack.a

The user can use ArpackNG_INCLUDE_DIRS and ArpackNG_LIBRARIES to compile/link with arpack-ng in his own CMakeLists.txt (after find_package).

If the user asks for an newer version, cmake breaks at configure time: this is checked by ArpackNGConfigVersion.cmake.


CMakeLists.txt
```diff
@@ -444,3 +444,13 @@ if (MPI)
       LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
       RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif ()
+
+# Provide FindArpackNG to users.
+configure_file(ArpackNGConfig.cmake.in "${PROJECT_BINARY_DIR}/ArpackNGConfig.cmake")
+install(FILES "${PROJECT_BINARY_DIR}/ArpackNGConfig.cmake" DESTINATION "${CMAKE_INSTALL_PREFIX}")
+set(ArpackNG_MAJOR_VERSION 3)
```
sylvestre: Is that mandatory to split MAJOR/MINOR/PATCH ?
why not having 3.5.0 in arpackng_version?

sylvestre:
please rename Arpack-NG to arpack-ng

CMakeLists.txt
fghoussen: No. I did like people usually do.

fghoussen:
Not sure to get what you mean. There is no Arpack-NG but only ArpackNG if I am not mistaken. This is camel case code style which is used usually in CMake projects (look at HDF5 or any project maintained by Kitware like ParaView for instance).

fghoussen:
I added a few lines in the CHANGE file for that too.
If you prefer I can also fixup these 2 commits into 1 for history clarity (let me know).
For now, I let them split for you to better see the patch I added according to your suggestion.

sylvestre:
Don't worry about the 2 commits into one. github fixes that
now, the project is called arpack-ng not Arpack-NG or ArpackNG. So, all references should be like the first one (sorry, I just value consistency)

fghoussen:
Done: is this what you expected ? I let flexibility for cmake users to call find_package with both flavours arpack-ng and/or ArpackNG (this depends on users !).

sylvestre:
I still don't understand why you want to call this project ArpackNG?
```
$ grep -r ArpackNG *|wc -l
0
$ grep -r Arpack-NG *|wc -l
0
$ grep -r arpack-ng *|wc -l
36

```

fghoussen:
I don't "want" anything in particular, I just follow the way it is usually done (from what I understand - I may be wrong). You can change/rename all that as you like: that's OK ! I just PR'ed to let you know about what matters: make things work !

sylvestre:
Merged and I fixed that here:
https://github.com/opencollab/arpack-ng/pull/71

fghoussen:
Note that this is not yet working for **p**arpack: know you may have a later issue about that.

I "only" handled arpack (start simple is better + only needed arpack for now) as I have "only" tested/used arpack so I only commit/pushed that (don't like blind commit/push of stuffs that "should" work without minimal testing). 

For parpack, same idea: need to complete arpack_ng_LIBRARIES with parpack but I don't remember if it's before or after arpack and/or if one need to add a " " or a ";" or a ":" between libs to add in arpack_ng_LIBRARIES.

Issue #64, opened 2017-07-03
cmake build offers no option to run EXAMPLES tests


rathann:
When arpack-ng 3.5.0 is configured using cmake, the tests in EXAMPLES are not run, whereas when using autotools, they are (72 more tests are run this way). Please update the CMakeFiles.txt to include running EXAMPLES tests, like configure'd Makefile does.

sylvestre:
@fghoussen As you are touching that, could you have a look?

fghoussen:
I already noticed that cmake provided check but no test (don't know neither why nor what the difference is supposed to be). check is OK (I modified that recently). I added a test target (see dedicated PR).

fghoussen:
BTW, this should be OK now ? If yes, I guess you can close this

sylvestre:
Sure

PR #74, opened 2017-10-26
Run the testsuite in the CI with cmake #64.

fghoussen:
Fix issue #64.
Cmake provided check but no test. Add a test target.


sylvestre:
Thanks!

PR #75, opened 2017-10-26
Issue 62: generate static/shared libraries.

fghoussen:


CMakeLists.txt
```diff
@@ -113,14 +113,13 @@ if (MPI)
...
-#                   ${arpackblas_STAT_SRCS}
-#                   ${arpacklapack_STAT_SRCS}
-#                   ${arpackutil_STAT_SRCS}
-#                   ${arpacksrc_STAT_SRCS})
-add_library(arpack ${arpackutil_STAT_SRCS}
-                   ${arpacksrc_STAT_SRCS}
-                   ${arpacksrc_ICB})
+option(ENABLE_STATIC "Static linking (shared otherwise)" ON)
+if (ENABLE_STATIC)
+  add_library(arpack STATIC ${arpackutil_STAT_SRCS} ${arpacksrc_STAT_SRCS} ${arpacksrc_ICB})
```
sylvestre: Can the STATIC/SHARED be moved into a variable to avoid duplicating the call to add_library() ?

CMakeLists.txt
```diff
@@ -133,10 +132,11 @@ set_target_properties(arpack PROPERTIES VERSION 2.0.0)
 set_target_properties(arpack PROPERTIES SOVERSION 2)
 
 if (MPI)
-#    add_library(parpack SHARED
-    add_library(parpack
-                        ${parpacksrc_STAT_SRCS}
-                        ${parpackutil_STAT_SRCS})
+  if (ENABLE_STATIC)
+    add_library(parpack STATIC ${parpacksrc_STAT_SRCS} ${parpackutil_STAT_SRCS})
```
sylvestre: ditto

CMakeLists.txt
fghoussen: Not sure... I would say no (STATIC/SHARED are keywords). You may wrap that in a cmake function.

CMakeLists.txt
```diff
@@ -113,14 +113,13 @@ if (MPI)
...
 
-#add_library(arpack SHARED
-#                   ${arpackblas_STAT_SRCS}
-#                   ${arpacklapack_STAT_SRCS}
-#                   ${arpackutil_STAT_SRCS}
-#                   ${arpacksrc_STAT_SRCS})
-add_library(arpack ${arpackutil_STAT_SRCS}
-                   ${arpacksrc_STAT_SRCS}
-                   ${arpacksrc_ICB})
+option(ENABLE_STATIC "Static linking (shared otherwise)" ON)
```
davydden: @sylvestre @fghoussen 

What's the rational in adding this option? [add_library](https://cmake.org/cmake/help/v3.0/command/add_library.html) already support the default `-D BUILD_SHARED_LIBS `:

> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON.

so IMO this PR should be reverted. This would also avoid code duplication @sylvestre mentioned below.

davydden: see https://github.com/opencollab/arpack-ng/pull/94

fghoussen:
>add_library already support the default -D BUILD_SHARED_LIBS

If so, I just didn't know this !... :D So you are probably right

PR #78, opened 2017-11-09
[CLEAN] CMake: kill useless lines (coming from bad merge ?).

fghoussen:
I believe these extra useless lines come from a bad merge.

Franck


sylvestre:
Thanks!

Issue #79, opened 2018-01-18
Incorrect largest magnitude eigenvalue


mhauru:
For the matrix in http://mhauru.org/trouble_matrix.zip, scipy.sparse.linalg.eigs, which calls ARPACK, gives the wrong dominant eigenvalue. The result depends on the initial guess for the eigenvector. Example python 3 code:
```
import sys
import numpy as np
import scipy as sp
import scipy.sparse.linalg as spsla

print(sp.__version__, np.__version__, sys.version_info)

M = np.loadtxt("trouble_matrix.out", dtype=np.complex_)
nev = 1
d = M.shape[0]

# Exact, full eig.
S, U = np.linalg.eig(M)
S = S[np.argsort(-np.abs(S))]
print(S[:nev])

# ARPACK, with different initial guesses.
v0 = np.random.rand(d)
S, U = spsla.eigs(M, k=nev, v0=v0)
print(S)
S, U = spsla.eigs(M, k=nev, v0=v0)
print(S)
v0 = np.random.rand(d)
S, U = spsla.eigs(M, k=nev, v0=v0)
print(S)

# Finding the dominant eigenvalue with a simple power iteration.
v = np.random.rand(d)
for _ in range(100):
    v /= np.linalg.norm(v)
    old_v = v
    v = np.dot(M, v)
print(np.average(v/old_v))
```
For fun, I've added a piece at the end that finds the correct dominant eigenvalue with a simple power iteration.

Output:
```
1.0.0 1.14.0 sys.version_info(major=3, minor=6, micro=3, releaselevel='final', serial=0)
[0.65419337-8.24548726e-17j]
[21.48733815-0.03849766j]
[21.48733815-0.03849766j]
[0.76760502-15.73229065j]
(0.654193372268301+1.8123127532736513e-16j)
```
Asking for more than one dominant eigenvalue goes equally wrong, I'm just choosing to showcase the simplest possible thing. I tried increasing ncv and maxiter, but to no help. In case it matters, the dominant eigenvalue of `M` is not degenerate.

I originally filed this as a scipy issue (https://github.com/scipy/scipy/issues/8307), but @pv checked the same matrix in Octave and ran into the same behavior, so presuming then that this is a problem with ARPACK.

I think I'm using ARPACK 3.3.0 since that's what comes with scipy 1.0.0.

pv:
Here's a smaller sparse test matrix, obtained by rounding and dropping small entries in the above:
[...]

PR #80, opened 2018-01-25
Force the residual vector to be in the range of OP. (#79)

caliarim:
Pull request for issue (#79) (force the initial residual to be in the range of the operator).

### DIFF
--- a/SRC/cgetv0.f
+++ b/SRC/cgetv0.f
@@ -239,24 +239,22 @@ c        | the generalized problem when B is possibly (singular).   |
 c        %----------------------------------------------------------%
 c
          call arscnd (t2)
-         if (bmat .eq. 'G') then
-            nopx = nopx + 1
-            ipntr(1) = 1
-            ipntr(2) = n + 1
-            call ccopy (n, resid, 1, workd, 1)
-            ido = -1
-            go to 9000
-         end if
+         nopx = nopx + 1
+         ipntr(1) = 1
+         ipntr(2) = n + 1
+         call ccopy (n, resid, 1, workd, 1)
+         ido = -1
+         go to 9000
       end if
 c 
 c     %----------------------------------------%
-c     | Back from computing B*(initial-vector) |
+c     | Back from computing OP*(initial-vector) |
 c     %----------------------------------------%
 c
       if (first) go to 20
 c
 c     %-----------------------------------------------%
-c     | Back from computing B*(orthogonalized-vector) |
+c     | Back from computing OP*(orthogonalized-vector) |
 c     %-----------------------------------------------%
 c
       if (orth)  go to 40
@@ -271,9 +269,9 @@ c     %------------------------------------------------------%
 c
       call arscnd (t2)
       first = .TRUE.
+      call ccopy (n, workd(n+1), 1, resid, 1)
       if (bmat .eq. 'G') then
          nbx = nbx + 1
-         call ccopy (n, workd(n+1), 1, resid, 1)
          ipntr(1) = n + 1
          ipntr(2) = 1
          ido = 2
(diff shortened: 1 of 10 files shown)


sylvestre:
Many thanks for your PR!

Seems that this is failing:
https://travis-ci.org/opencollab/arpack-ng/jobs/333185780
The following tests FAILED:
	 77 - bug_79_double_complex (Not Run)

caliarim:
I forgot to add the test to add_dependencies in CMakeLists.txt?

sylvestre:
@turboencabulator  @thrasibule
any option on this?

turboencabulator:
Build script changes look good to me.

SRC/cgetv0.f
```diff
@@ -271,9 +269,9 @@ subroutine cgetv0
 c
       call arscnd (t2)
       first = .TRUE.
+      call ccopy (n, workd(n+1), 1, resid, 1)
```
thrasibule: Why is this needed? We only end up here if ido != 0.

thrasibule:
This seems pretty low risk, and it fixes a test case, so looks good to me.

SRC/cgetv0.f
caliarim: I'm away from my laptop, but aftet the very first OP application ido is -1, isn't it? So we need to copy the result of OP application back to resid.

caliarim:
Shall we change the documentation somewhere about this change? For instance, in X{s,n}aupd it is written

> c  RESID   Complex*16  array of length N.  (INPUT/OUTPUT)
> c          On INPUT:
> c          If INFO .EQ. 0, a random initial residual vector is used.
> c          If INFO .NE. 0, RESID contains the initial residual vector,
> c                          possibly from a previous run.

No mention to the range of the operator. Basically the same in X{s,n}aup2. In Xgetv0 it is written

> c  Force the residual vector to be in the range of the operator OP.

In the book, page 72, it is written

> In either case, it the (shift-invert) computational mode calls for it, the vector is also forced to be in the range of OP.

But with this change the vector is forced to be in the range for the regular mode, too. And it has always been forced to be in the range for the generalized problem.
And what happens if somebody really would like to test her/his own inital vector?

sylvestre:
@caliarim is that ready to be merged? Thanks

caliarim:
I think so. With respect to my previous comment, in Xgetv0 it is written that the initial vector is in the range, and now this is (always) true. And of course we cannot change the book...

sylvestre:
thanks!

ViralBShah:
Apologies for chiming in on such an old issue. We believe this PR is causing the Arpack tests in Julia to fail. The Julia test case is described in https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/118, and fails once in a while when run repeatedly.

It may very well be an issue in our interface, but this worked reliably in Arpack 3.5, and not in any of the other later releases. In case someone here notices something we are doing wrong, or has some insight - it would be great.

ViralBShah:
Apologies for chiming in on a closed PR. For Julia's Arpack.jl, we believe that this has introduced a bug (it is our best guess):

The test case in https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/118 demonstrates the following behaviour:
1. Fails about 2% of the time with arpack-ng 3.5. If initialized with a random `v0`, it passes reliably.
2. Fails always on later versions of arpack-ng, including 3.8, even when initialized with a random `v0` (demonstrated in https://github.com/JuliaLinearAlgebra/Arpack.jl/pull/145)

sylvestre:
@ViralBShah nice to see you there and no worries for this.

@caliarim do you want to propose a PR ? :)
thanks

fghoussen:
@ViralBShah: could revert both d04bdf19b9189a971f8c3bf2561b8058994e9cb4 and 31854cadaff067e30b8715f10533c459a9b7d447 related commits and check things are getting better at your side? Making sure the reg you see comes from that

AFAIU, this would imply reopening #79 and #142!...

PR #84, opened 2018-02-13
Fix C and C++ bindings ABI

### LINKED ISSUE #83: Wrong type corresponding to arrays of `logical` in arpack.hpp
For `dseupd_c` we have `bool * select` in `arpack.hpp`, which, if really passed an buffer `bool[ncv]`, will result in buffer overflow, because (on Linux x86) `sizeof(bool)==1` while size of Fortran's `logical` is 4.
It seems actual type for this pointer should be `int*` or something like this.
The same might be true for `arpack.h`.

10110111:
Fortran's logical type has size of 4 bytes, while C and C++ have `sizeof(bool)==1`. Thus it's no good to pass `bool[ncv]` array to a function expecting `logical(ncv)`.
Currently attempt to do this results in successful compilation and subsequent buffer overflow in e.g. `dseupd_c`. This is actually caught at run time by gcc>=5 by its stack smashing detector, leading to test failures.
This patch replaces `bool*` types expected by C and C++ bindings with `int*`, and corresponding declarations of the arrays in callers in the tests from `bool[3*ncv]` to `int[ncv]`.

### DIFF
--- a/TESTS/icb_arpack_c.c
+++ b/TESTS/icb_arpack_c.c
@@ -43,7 +43,7 @@ int ds() {
   bool rvec = true;
   char howmny[] = "A";
   double* d = (double*) malloc((nev+1)*sizeof(double));
-  bool select[3*ncv];
+  int select[ncv];
   double z[(N+1)*(nev+1)];
   BLASINT ldz = N+1;
   double sigma=0;
@@ -110,7 +110,7 @@ int zn() {
   bool rvec = true;
   char howmny[] = "A";
   double _Complex* d = (double _Complex*) malloc((nev+1)*sizeof(double _Complex));
-  bool select[3*ncv];
+  int select[ncv];
   double _Complex z[(N+1)*(nev+1)];
   BLASINT ldz = N+1;
   double sigma=0;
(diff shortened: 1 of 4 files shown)


sylvestre:
As this is changing the API/ABI, I am afraid I cannot take this patch :( 
sorry

10110111:
But without this patch the ABI is already broken. I discovered #83 exactly due to this: I was trying to follow documentation and got segmentation faults and aborts (depending on what got overwritten).
And once you try `make check` with default compiler being gcc>=5, you'll get test failures for the same reason.

sylvestre:
OK, could you please add a test which shows the issue?

(please also update the CHANGES file)

10110111:
A way to demonstrate the problem with current API is
1. allocate e.g. `4*NCV*sizeof(ExpectedBooleanType)` bytes, where `ExpectedBooleanType` is the type the functions like `dseupd_c` expect for `select` parameter
2. fill them with some recognizable pattern
3. pass them as `select` parameter to `dseupd_c`
5. check whether any bytes beyond `NCV*sizeof(ExpectedBooleanType)` are changed
6. If any are, we indeed have a problem

But this doesn't provide any means of checking that `ExpectedBooleanType` indeed matches what the function expects, unless we check the whole function signature. More or less easy way to get type of `N`th parameter seems to exist in C++11, but `icb_arpack_cpp` test doesn't use it, so I'm not sure whether I can introduce this requirement.

Any suggestions?

thrasibule:
arpack.h and arpack.hpp have never been part of a released version, and it's not changing the ABI, just the headers, which were incorrect. I think the fix looks good.

sylvestre:
thanks!

Issue #86, opened 2018-03-07
C/C++ bindings const and std::complex


dbeurle:
I've got a couple of questions regarding the C/C++ bindings and examples for arpack-ng.

Firstly, is it possible to make the C/C++ bindings const correct?  This would help signal what changes when passing in char pointers, or better yet accept a `std::string` and use `c_str()`.

Secondly, the C++ example uses `_Complex` (C style) and most C++ codes use the standard library implementation of `std::complex<>`.  Is it possible to change the bindings and example to accept a std::complex<>* instead of the C style pointer to _Complex?  It's possible to cast this back to an array with the same object representation for C compatibility: http://en.cppreference.com/w/cpp/numeric/complex.

I have spent some spare time improving the C++ example and I'm happy to help with these changes.

sylvestre:
@10110111 @fghoussen 
You might be interested by this proposition.

Sure, const would be good!

dbeurle:
Const is the best :+1:

10110111:
Indeed, I'm all for const-correctness. Actually I was thinking of proposing the change after my previous fix was accepted.

As for `_Complex` vs `std::complex`, they are layout-compatible, so I suppose simply changing the header to use `std::complex` in the C++ version should work without extra work in the Fortran side of the binding.

But to accept `std::string` instead of `const char*` you'd have to wrap the original Fortran function into a converter function. To me it doesn't look too useful to be worth it, especially since it may involve useless dynamic allocation for very short (1-2 char) strings.

dbeurle:
I think SSO will avoid heap allocation here but if I can't convince you then keeping it as `char const*` is ok (I really dislike pointers though).  A safer approach would be a strongly typed enum and convert that to `char[]` so the user really can't make an error with the casing.  This of course means adding in some converter functions to the C++ bindings.

For now though, if I changed the C and C++ bindings to be `const` correct and the C++ bindings to use `std::complex<T>` are you available to review this PR?  Then the examples can be updated to show off the new bindings.

10110111:
An enum is the best option, and for this one IMO it's worth making converter functions.

sylvestre:
Not sure about c++11 ...
Users of fortran libs have usually a dependency to use old and deprecated compilers...
However, if this is making your life way easier, please go ahead (and add a checks in the configure and cmake)

fghoussen:
[...]

PR #87, opened 2018-03-07
Test PR

dbeurle:
Test PR on changes to the test script for the c++ arpack example.  Related to #86

### DIFF
--- a/TESTS/icb_arpack_c.c
+++ b/TESTS/icb_arpack_c.c
@@ -90,7 +90,7 @@ int ds() {
 void zMatVec(double _Complex * x, double _Complex * y) {
   int i;
   for (i = 0; i < 1000; ++i)
-    y[i] = (i+1.)*creal(x[i]) + I * (i+1.)*cimag(x[i]);
+    y[i] = x[i] * (i+1.0 + _Complex_I * (i+1.0));
 };
 
 int zn() {
@@ -147,7 +147,7 @@ int zn() {
   int i;
   for (i = 0; i < nev; ++i) {
     printf("%f %f\n", creal(d[i]), cimag(d[i]));
-    if(fabs(creal(d[i]) - (double)(1000-i))>1e-6){
+    if(fabs(creal(d[i]) - (double)(1000-i))>1e-6 || fabs(cimag(d[i]) - (double)(1000-i))>1e-6){
       free(d);
       return 1;
     }
--- a/TESTS/icb_arpack_cpp.cpp
+++ b/TESTS/icb_arpack_cpp.cpp
@@ -89,7 +89,7 @@ int ss() {
 void cMatVec(float _Complex * x, float _Complex * y) {
   int i;
   for (i = 0; i < 1000; ++i)
-    y[i] = (i+1.)*creal(x[i]) + I * (i+1.)*cimag(x[i]);
+    y[i] = x[i] * (i+1.0f + _Complex_I * (i+1.0f));
 };
 
 int cn() {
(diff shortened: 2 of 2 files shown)


dbeurle:
@sylvestre Can I get a hand please with the configure build script?  I'm not familiar with autotools (millennials eh?).  Best I could find is https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html

I guess `noext` and `required` are the options?

fghoussen:
You should not need C++11.
Why do you rewrite the all example ? This makes history messy and loose the reviewer: you just need to replace ssaupd_c (and others) with the new C++ wrappers. There is only a few lines to change.

dbeurle:
Do you think that example is clear and easier to follow?  For example, allocating memory on the stack and heap inconsistently, casting the result from `new` and is plain confusing to a newcomer.  The example should be best practise.  You want to people to understand what is happening in the test / example code and what was there before was difficult to read and outdated.  There is no point having a C++ example if all that is different is a call to `new` and `malloc`.

fghoussen:
This example comes from the C example (not meant for style, meant for checking it works).
I meant that to keep history clean/understandable by all people involved, it's better to stick to minimal changes only (personal opinion).

dbeurle:
@fghoussen 
> This example comes from the C example (not meant for style, meant for checking it works).
What exactly is being checked in the C++ example?  Whether `new` or `extern "C"` works?  The C and C++ tests don't even check if the imaginary eigenvalue is correct (which they aren't on my machine).  The examples for the complex matrices are broken.

The output of an octave script
```matlab
A = sparse(diag(1:10) + I * diag(1:10)); [ei, ev] = eigs(A)
```
gives the expected result - which IIRC references ARPACK on the backend.

> I meant that to keep history clean/understandable by all people involved, it's better to stick to minimal changes only (personal opinion).

My personal opinion is that history is just that.  If that example file needs to be rewritten to be clear and show good use of the API then one commit is a price I'd be willing to pay.  If the maintainers disagree then so be it.  I'm not trying to create a mess - I'm trying to raise the quality of the C++ examples.  You can pooh pooh me at every step or you can admit it needs improvement and we can work together to make this happen.

fghoussen:
I understand your intention which is commendable. Arpack has been written in the 70's by cavern guys with the ugly permissive fortran : it's no more maintained and has been a pain to use by everybody from the very beginning, nobody will change that now ! iso_c_binding is just a gateway: no more. There is no place for style or quality here (I wish it could). Arpack would need a top-down full re-write in modern C++ (replacing the permissive and misleading Reverse Communication Interface with modern C++ std::function) but it's so specific, tricky and time-consuming that nobody never did it. There is no way now to make arpack user-friendly (personal opinion). Same story for LAPACK, SCALAPACK and others...

fghoussen:
Don't want to discourage you here. I tried to PR something...

dbeurle:
I agree, the interface is horrible but the other options aren't particularly attractive either (unless you know of a good eigenvalue library?).  A google search for C++ and sparse eigenvalue solvers returns SLEPc (need to pull in PETSc), FEAST which requires sign up, ARPACK (that's why I'm here) and Spectra (doesn't seem to do complex matrices).  If there were another library that was easier to use and as well tested then other projects would switch.  That's probably why does this project exists.

That `const` PR you did is a step in the right direction in making this interface better.  At some point we're putting lipstick on a pig, I agree, but until someone wants to pony up and write this modern eigenvalue solver we're stuck with this.

Are you familiar enough with autotools to make the `c++11` check?

fghoussen:
>What exactly is being checked in the C++ example? Whether new or extern "C" works?

Just that the C example can be compiled / link with a C++ compiler. No more.

>the interface is horrible but the other options aren't particularly attractive either 

With *PACK (LAPACK, SCALAPACK, ARPACK, ...) the interface always as cumbersome as depressive : you will be _compelled_ to live with these painful things as there is no other options... Style is out of question.

>search for C++ and sparse eigenvalue solvers

SLEPc (which uses ARPACK) is not so bad. You can check out eigen (maybe not everything is available - don't know for complex) : http://eigen.tuxfamily.org/index.php?title=Main_Page.

>That const PR you did is a step in the right direction in making this interface better.

You have no chance to have better than this. For std::complex, actually you need std to provide std::complex::c_cplx (the way std::string::c_str does)... Good luck on that !...

>Are you familiar enough with autotools to make the c++11 check?

Use ax_cxx_compile_stdcxx_11

PR #91, opened 2018-03-10
Test imaginary eigenvalue

dbeurle:


dbeurle:
@fghoussen

sylvestre:
I guess you are aware that the tests failed.

dbeurle:
Exactly as I planned :+1:

fghoussen:
Remember arpack has been written by cavern guys in the 70's : it is likely that data are not initialized the right way and arpack does not care or tell even if he knows... This would be typical a cavern's coding style/rules... 

I had yank-pasted the ds/ss versions of the C example to write the zn/cn ones (trying to test more functions of the API - testing all of them was too tedious...). Input data are probably not initialized the right way (I probably missed that). For a cavern guy, this is not a bug : this is a user misuse.

I would say the workd/l initialisation for float/double _Complex zeroes only the real part... So the imaginary part is "obviously" wrong (for cavern's). If so, be sure that there is no way to fix this in arpack (not maintained + cumbersome), you'll need to change the initialization in the examples (which would be a good thing). Let your test which is a good thing to keep, and, Initialize workd/l (in the zn and cn cases only) like this:
`workd/l[k] = 0. + I * 0.;`
in the examples, commit, and update your PR with that...

dbeurle:
Explains the result and demonstrates the disadvantages of C, `std::complex<>` default constructs both members to zero which explains why it worked.  I'll update the C and C++ tests.

fghoussen:
Can you update the tests to see if this works ?

dbeurle:
It doesn't work - `workl` looks like `0.0f + 0.0f * I` on initialisation.

fghoussen:
Oh men !... I guess I got it. Really need to leave now. Should be able to PR tomorrow a fix for this.

dbeurle:
@sylvestre Good to go.

sylvestre:
Thanks, looks good!

fghoussen:
Yes ! :D This is what I had in mind yesterday evening... And was in a rush to leave. Perfect !

PR #94, opened 2018-03-12
cmake: use standard BUILD_SHARED_LIBS to control static vs shared

davydden:


sylvestre:
As you are changing a build argument, could you please at least mention that in the changelog?
Being able to handle the previous argument would have been great too.

davydden:
>  could you please at least mention that in the changelog?

there is no need, this was never released. I am just restoring the behaviour of the currently stable release `3.5.0` which **already supprted static/shared builds**.

> Being able to handle the previous argument would have been great too.

As I explained in the [PR which introduced this option](https://github.com/opencollab/arpack-ng/pull/75), [add_library](https://cmake.org/cmake/help/v3.0/command/add_library.html) **already** supports the default `-D BUILD_SHARED_LIBS`:

> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON.

I think the option was added because the author was not aware of this behaviour of the `add_library`.

sylvestre:
ok, merci!

fghoussen:
>I think the option was added because @fghoussen was not aware of this behaviour of the add_library.

Yes, I didn't know that

PR #95, opened 2018-03-14
parpack : add iso_c_binding support.

fghoussen:
The code and cmake tests work on debian/testing.

There is a cmake specific problem on travis/ubuntu : travis fails (icb_parpack_cpp link KO) => if somebody can help here...
Not a autotools guru : autotools tests do not compile / run => if somebody can help here...

If somebody can help : would be good, feel free to commit at my side (https://github.com/fghoussen/arpack-ng/commits/icb_parpack)... And otherwise no such a big deal : forget the PR and close this.

Franck

### DIFF
--- a/parpack.hpp
+++ b/parpack.hpp
@@ -1,0 +1,85 @@
+#ifndef __PARPACK_HPP__
+#define __PARPACK_HPP__
+
+/*
+ * From C++, arpack does not exist.
+ * Arpack is Fortran. ISO_C_BINDING is a gateway from Fortran to C, not C++. But, C++ can "get back" to C.
+ * This is why you find C types in the arpack.hpp (C++ header) to be as safe as possible.
+ */
+
+/*
+ * IMPORTANT: MPI communicators MUST be passed from C to Fortran using MPI_Comm_c2f.
+ *            MPI_Fint MCW = MPI_Comm_c2f(MPI_COMM_WORLD);
+ */
+
+#include "mpi.h"
+
+extern "C" void pssaupd_c(MPI_Fint comm, int & ido, const char * bmat, int n, const char * which, int nev,
+                          float tol, float * resid, int ncv, float * v,
+                          int ldv, int * iparam, int * ipntr, float * workd,
+                          float * workl, int lworkl, int & info);
+
+extern "C" void psseupd_c(MPI_Fint comm, bool rvec, const char * howmny, int * select, float * d, float * z, int ldz, float sigma,
+                          const char * bmat, int n, const char * which, int nev,
+                          float tol, float * resid, int ncv, float * v,
+                          int ldv, int * iparam, int * ipntr, float * workd,
+                          float * workl, int lworkl, int & info);
+
+extern "C" void pdsaupd_c(MPI_Fint comm, int & ido, const char * bmat, int n, const char * which, int nev,
+                          double tol, double * resid, int ncv, double * v,
+                          int ldv, int * iparam, int * ipntr, double * workd,
+                          double * workl, int lworkl, int & info);
+
+extern "C" void pdseupd_c(MPI_Fint comm, bool rvec, const char * howmny, int * select, double * d, double * z, int ldz, double sigma,
+                          const char * bmat, int n, const char * which, int nev,
+                          double tol, double * resid, int ncv, double * v,
+                          int ldv, int * iparam, int * ipntr, double * workd,
+                          double * workl, int lworkl, int & info);
+
+extern "C" void psnaupd_c(MPI_Fint comm, int & ido, const char * bmat, int n, const char * which, int nev,
+                          float tol, float * resid, int ncv, float * v,
+                          int ldv, int * iparam, int * ipntr, float * workd,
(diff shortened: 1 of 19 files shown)


fghoussen:
cmake was actually OK on both debian and ubuntu. The problem was in travis.yml : you need to rm CMakeCache.txt at least or the next cmake is polluted with previous settings... So stupid waste of time... Anyway

I turned off the autotools icb tests : I let that for guru's... (updated the TODO and added TODO flags)

The code (iso_c_binding) is working and tested with cmake (not autotools - TODO).

I let that this way: I will not have mo time for this.

Take it or leave it !...

fghoussen:
@sylvestre 
By the way, what do you say ? I could kill the autotools commit if you prefer (so cmake would test all but autotools would not) ?

Really sorry about that but the autotool part is really cubersome: I tried to "mimic" the hdf5's autotools but impossible to get it to work !...

fghoussen:
... Or what about dropping autotools, now that you have cmake ? would probably be easier to maintain (personal opinion)

sylvestre:
If you document the issue in the changelog, I will merge it.
thanks again for your work!

sylvestre:
About killing the autotools, this is a different topic... 
Please open an issue to start the discussion on that!

fghoussen:
>If you document the issue in the changelog, I will merge it.

I pushed a different version of the very last commit (autotools) which seems more logical / minimal to me. I tried to describe the autotools problem and give hints in the TODO file...

>About killing the autotools, this is a different topic...

Sure ! I am not asking to kill autotools !... But when it comes to writing m4 macros (like HDF5 did to handle icb), it becomes generally more and more cryptic :D I know basics, but, I am not a guru. I guess dropping autotools would make the package easier to maintain, but, I can understand you want to keep it.

parpack.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+/*
+ * IMPORTANT: MPI communicators MUST be passed from C to Fortran using MPI_Comm_c2f.
+ *            MPI_Fint MCW = MPI_Comm_c2f(MPI_COMM_WORLD);
+ */
+
+#include "mpi.h"
+
+extern "C" void pssaupd_c(MPI_Fint comm, int & ido, const char * bmat, int n, const char * which, int nev,
```
davydden: maybe have a single header and just do
```
#ifdef __cplusplus
extern "C" {
#endif
  // blabla

#ifdef __cplusplus
}
#endif
```
? Maintaining two files with identical definition is probably not ideal

fghoussen: Sure. For now I was fighting (!) for having a very first version that works ! :D 

When this is merged. I guess it can be a good thing to do what you propose for both arpack (.h, .hpp) and parpack (.h and .hpp)

davydden: i just wanted to make sure that this happens before the release, otherwise you won't be able to change it in a backward-compatible way.

fghoussen: I understand. This can be done quick and fast. I'll PR that after this PR is merged.

@sylvestre 
Is this PR OK for you ? (as it is now)

@davydden 
Just to make sure we are OK: you propose to kill *arpack.hpp and to have only *arpack.h that could be included both from C or C++ ? Right ? I am fine with that.
For now, the only thing I fighted for was really to get working !
The other option would be to include *arpack.h in *arpack.hpp (not tested yet but I guess this could work) and may be more "C++ oriented" (I am fine with that too - this is only a question of style, so it's no big deal to me - to me, what's important is that it works ! :D)

sylvestre: More or less... I wish you had some docs or tutorial to explain how to use it.
but I won't block on it.
Please let me know when this can be merged!

fghoussen: I've added a README as a doc/example (like I did for arpack). For me you can merge this.

I'll try to PR quickly what @davydden  proposes this afternoon or tomorrow if I can access the web.

davydden: > you propose to kill *arpack.hpp and to have only *arpack.h that could be included both from C or C++ ? 

correct.

sylvestre: It isn't what I call documentation but sure :p

davydden: @fghoussen if you create a PR, it would be easier to see

fghoussen: > It isn't what I call documentation but sure :p

Documenting arpack is "following the examples" !... :D 
For icb, this is the only one way to go (like hdf5) : it's commented in the .c and .cpp files (the README redirect to them).

davydden: > For icb, this is the only one way to go

[doxygen](http://www.stack.nl/~dimitri/doxygen/)?

fghoussen: >doxygen ?

Too much of a burden ! I am really running out of time for this... Can't afford that. But I let you do that if you feel this is necessary

>@fghoussen if you create a PR, it would be easier to see

PR'ed !

davydden: > Too much of a burden !

fully agree, i just wanted to mention that other options.

> But I let you do that if you feel this is necessary

certainly not necessary enough to commit to it ;-)

PR #98, opened 2018-03-21
iso_c_binding: kill hpp (get one header only for both C/C++).

fghoussen:


fghoussen:
Feel free to commit here : https://github.com/fghoussen/arpack-ng/commits/icb as you like !

fghoussen:
My understanding is that extern is mandatory for C to "find" Fortran : this is where it seems to break.

The API is frozen : once done, having both a .h and a .hpp files will not trigger any more problems !... Not sure this PR is really necessary.

I let maintainers decide what to do (I understand both points of view - keeping or kill hpp files). You decide !

sylvestre:
I would prefer having only one file.

Is there really people would use it only in c mode?
Every compiler now has C and C++ mode...

davydden:
> I would prefer having only one file.

but you do have a single file for serial `arpack` and another single for `parpack`.

sylvestre:
I prefer an header per feature but I won't mind having one for the two.

fghoussen:
>for pure C. But that's just my lack of knowledge about pure C

My understanding is that for pure C extern is mandatory (it tells C that the function is "elsewhere" that is in Fortran - without it, I guess compilation is OK but link is KO). Feel free to test/PR at my side.

For header(s): they will be frozen anyway !...

davydden:
> Feel free to test/PR at my side.

(un)fortunately i don't use `C` at all, but `C++` side certainly LGTM.

davydden:
@fghoussen i googled and I think you are right that `extern` is needed for `C`.

davydden:
@fghoussen you can drop the last commit, the first one alone should indeed be correct.

fghoussen:
not sure to understand. the first commit was breaking CI anyway. 
for me (not 100% sure) extern is needed in pure C so this will be a problem if you want to suppress it.

davydden:
> not sure to understand. the first commit was breaking CI anyway.
for me (not 100% sure) extern is needed in pure C so this will be a problem if you want to suppress it.

when I commented, you had two commits, where the second one was removing `extern` in pure `C`. 

I do agre with you

> extern is needed in pure C

and in the current state this should be all good.

fghoussen:
>and in the current state this should be all good.

The current state is broken. Not sure to know how to get that to work

davydden:
@fghoussen indeed
```
[ 83%] [32mBuilding C object CMakeFiles/icb_parpack_c.dir/PARPACK/TESTS/MPI/icb_parpack_c.c.o[0m
In file included from /home/travis/build/opencollab/arpack-ng/PARPACK/TESTS/MPI/icb_parpack_c.c:14:0:
/home/travis/build/opencollab/arpack-ng/parpack.h:14:8: error: expected identifier or â€˜(â€™ before â€˜{â€™ token
 extern {
        ^
make[2]: *** [CMakeFiles/icb_parpack_c.dir/PARPACK/TESTS/MPI/icb_parpack_c.c.o] Error 1
make[1]: *** [CMakeFiles/icb_parpack_c.dir/all] Error 2
make: *** [all] Error 2

travis_time:end:015af46e:start=1521727968354458072,finish=1521728009730268595,duration=41375810523
[0K
[31;1mThe command "if [ "$BUILD" == "cmake" ]; then make all check ; fi" exited with 2.[0m
travis_time:start:2a028a28
[0K$ if [ "$BUILD" == "cmake" ]; then make test ; fi
[36mRunning tests...[0m
Test project /home/travis/build/opencollab/arpack-ng/build
```

hm, what happens if you do the following construct
```
#ifdef  __cplusplus
extern "C" {
#endif
...
extern <blabla>
...
#ifdef  __cplusplus
}
#endif
```
?

fghoussen:
Yes, I knew about the error.
Already tried what you suggest : does not work, extern must appear once only (OK for C, KO for c++).

fghoussen:
For me, this PR can be closed. I don't see a solution to merge headers... I let you decide what to do

davydden:
> I let you decide what to do

by "you" you certainly mean @sylvestre  😄 

p.s. I contributed just a couple of minor fixes :)

fghoussen:
No, by "you" I mean "anybody" ! :D I do believe that arpack is a painful thing (should be fully re-written...) and that arpack-ng is a very courageous initiative (that's why I tried to help)

sylvestre:
To be honest and clear, I am just maintaining it because it is used by so many projects and someone has to do it...
I am not even using it myself. but for now, I haven't break anything (afaik).
I am just being the guy who accepts/rejects PR and make releases from time to time.

fghoussen:
I guess this can be closed too !... As it seems there is no simple solution for that.

PR #99, opened 2018-03-22
parpack: running autotools tests.

fghoussen:
note: ax_mpi must be ran twice (first run in --enable-mpi scope
is lost when entering --enable-icb scope).


fghoussen:
To me, this may not solve everything but it's a step in the right direction.

PR #104, opened 2018-03-30
Coverage by codecov via github.

fghoussen:
If you subscribe to codecov (free), you can have coverage on github push.
Check here : https://github.com/fghoussen/arpack-ng/commits/cov (click on green cross of 4fff365).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,47 @@
+codecov:
+
+  branches: *
+
+  ci:
+
+    - travis-ci.org
+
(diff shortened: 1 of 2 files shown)


sylvestre:
I added support for coveralls last week. Not sure what is the point of having the two?!

fghoussen:
Didn't know coveralls (I usually use codecov) !... I believed it was the name of the cmake option you added ! :D You are right : there is no point, I close this.

PR #106, opened 2018-04-05
CMake: cmake files install.

fghoussen:
find_package files (*.cmake) stand for pkg-config files (*.pc).
pc files are supposed to be installed in local/lib/pkgconfig.
cmake files are supposed to be installed in local/lib/cmake.


sylvestre:
What about adding that to the changelog?

fghoussen:
Already in changelog, line 8:
 * Provide find_package for (cmake) users to find arpack-ng.

No ?... Or I missed what you mean

sylvestre:
ok, my bad.

Issue #82, opened 2018-02-07
COMMON block variables are not initialized in pArpack


davydden:
I have some errors on HPC cluster (don't see anything like that on macOS or Ubuntu 16 or 17):

```
At line 68 of file <blabla>/arpack-ng-3.4.0/PARPACK/UTIL/MPI/pdvout.f
Fortran runtime error: Bad unit number in statement
```

The same issue with `3.5.0`.

From the description, it seems that the routine is for printing https://github.com/opencollab/arpack-ng/blob/master/PARPACK/UTIL/MPI/pdvout.f#L68 

but I don't see where/why Arpack would print anything while solving symmetric hermitian generalized eigenproblem with shift-and-invert spectral transformation.

EDIT: I googled and it looks like COMMON variables do not have default initialization. Since I call arpack from C++ and link against both Arpack and pArpack, for now my workaround of this problem is
```
sed -i -e 's/\(\s*msglvl = \)\(.*\)/\10/g' *.f
```

caliarim:
You have msglvl .gt. 0 somewhere.

davydden:
what is `msglvl` ?

caliarim:
In debug.h there are several integers which can be used to get information from the subroutines. For instance, if msaupd is 1 in your main program, then pdsaupd.f will set msglvl to msaupd and print some information. The output is redirected to unit logfil, which should be set to 6 (stdout).

davydden:
@caliarim sorry, I am lost here as I don't really work with Fortran much. I am using C-wrappers to pArpack functions from a C++ application. There is indeed `debug.h` in the `SRC` file, but my `pArpack` installation has no headers. It's probably missing in `cmake`'s scripts.

Also I do not set `msaupd` to anything as I was simply not aware of it. I would expect that the default is no debug output? You probably refer to this docu http://www.caam.rice.edu/software/ARPACK/UG/node133.html ? 

p.s. It would be good if one could disable all output at configure time with `cmake`.

davydden:
@caliarim 

> if msaupd is 1 in your main program, then pdsaupd.f will set msglvl to msaupd and print some information

could you please help clarify this given my previous reply? Especially how is it possible that this output happens if I call `parpack` wrappers from C++ and I certainly don't set `msaupd`?

I tried to check default values of those variables within (p)Arpack sources, but did not find anything but a few examples/tests where `msaupd` is set. I am not a Fortran user, but I guess the default is zero, so technically I should not see any output.

[...]

PR #107, opened 2018-04-05
iso_c_binding: add debug support.

fghoussen:
This should fix issue #82

### DIFF
--- a/debug_icb.f90
+++ b/debug_icb.f90
@@ -37,16 +37,3 @@ subroutine debug_c(logfil_c, ndigit_c, mgetv0_c,                                
   mcgets = mcgets_c
   mceupd = mceupd_c
 end subroutine debug_c
-
-! Initialisation of the debug common block to "no debug".
-block data init
-common /debug/ logfil, ndigit, mgetv0,                                &
-               msaupd, msaup2, msaitr, mseigt, msapps, msgets, mseupd,&
-               mnaupd, mnaup2, mnaitr, mneigh, mnapps, mngets, mneupd,&
-               mcaupd, mcaup2, mcaitr, mceigh, mcapps, mcgets, mceupd
-data logfil, ndigit, mgetv0,                                &
-     msaupd, msaup2, msaitr, mseigt, msapps, msgets, mseupd,&
-     mnaupd, mnaup2, mnaitr, mneigh, mnapps, mngets, mneupd,&
-     mcaupd, mcaup2, mcaitr, mceigh, mcapps, mcgets, mceupd &
-     / 6, -3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 /
-end
--- a/debug_icb.f90
+++ b/debug_icb.f90
@@ -1,0 +1,39 @@
+! iso_c_binding : debug
+
+subroutine debug_c(logfil_c, ndigit_c, mgetv0_c,                                        &
+                   msaupd_c, msaup2_c, msaitr_c, mseigt_c, msapps_c, msgets_c, mseupd_c,&
+                   mnaupd_c, mnaup2_c, mnaitr_c, mneigh_c, mnapps_c, mngets_c, mneupd_c,&
+                   mcaupd_c, mcaup2_c, mcaitr_c, mceigh_c, mcapps_c, mcgets_c, mceupd_c)&
+                   bind(c, name="debug_c")
+  use :: iso_c_binding
+  implicit none
+  integer(kind=c_int), value, intent(in) :: logfil_c, ndigit_c, mgetv0_c
+  integer(kind=c_int), value, intent(in) :: msaupd_c, msaup2_c, msaitr_c, mseigt_c, msapps_c, msgets_c, mseupd_c
+  integer(kind=c_int), value, intent(in) :: mnaupd_c, mnaup2_c, mnaitr_c, mneigh_c, mnapps_c, mngets_c, mneupd_c
+  integer(kind=c_int), value, intent(in) :: mcaupd_c, mcaup2_c, mcaitr_c, mceigh_c, mcapps_c, mcgets_c, mceupd_c
+  include 'debug.h'
+  logfil = logfil_c
+  ndigit = ndigit_c
+  mgetv0 = mgetv0_c
+  msaupd = msaupd_c
+  msaup2 = msaup2_c
+  msaitr = msaitr_c
+  mseigt = mseigt_c
+  msapps = msapps_c
+  msgets = msgets_c
+  mseupd = mseupd_c
+  mnaupd = mnaupd_c
+  mnaup2 = mnaup2_c
+  mnaitr = mnaitr_c
+  mneigh = mneigh_c
+  mnapps = mnapps_c
+  mngets = mngets_c
+  mneupd = mneupd_c
+  mcaupd = mcaupd_c
+  mcaup2 = mcaup2_c
+  mcaitr = mcaitr_c
+  mceigh = mceigh_c
+  mcapps = mcapps_c
+  mcgets = mcgets_c
+  mceupd = mceupd_c
+end subroutine debug_c
--- a/debug_icb.f90
+++ b/debug_icb.f90
@@ -37,3 +37,16 @@ subroutine debug_c(logfil_c, ndigit_c, mgetv0_c,                                
   mcgets = mcgets_c
   mceupd = mceupd_c
 end subroutine debug_c
+
+! Initialisation of the debug common block to "no debug".
+block data init
+common /debug/ logfil, ndigit, mgetv0,                                &
+               msaupd, msaup2, msaitr, mseigt, msapps, msgets, mseupd,&
+               mnaupd, mnaup2, mnaitr, mneigh, mnapps, mngets, mneupd,&
+               mcaupd, mcaup2, mcaitr, mceigh, mcapps, mcgets, mceupd
+data logfil, ndigit, mgetv0,                                &
+     msaupd, msaup2, msaitr, mseigt, msapps, msgets, mseupd,&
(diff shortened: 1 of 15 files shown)


fghoussen:
CI was OK at my side ?! No time for this today, I'll try to fix / check it this week-end. Would say if you dont call debug_c, there is no debug (init already OK). Not sure. Will check

davydden:
> Would say if you dont call debug_c, there is no debug (init already OK)

Form my experience, no. Note that you can't easily check this as https://github.com/opencollab/arpack-ng/issues/82 happened on HPC cluster whereas I also don't have any problems / debug output on my mac or Ubuntu. 
As I mentioned in https://github.com/opencollab/arpack-ng/issues/82 , simply patching source to disable all output whatsoever fixed it on a cluster.

fghoussen:
CI is OK at my side !?... Weird. I'll check it when possible

fghoussen:
Would say CI fails as coverage go from 70% to 69% ! :D I'll check tomorrow

fghoussen:
I've pushed comestic changes that should illustrate things the way you (@davydden) probably want them to be : hope this will get the coverage back from 69.5 to 70% !... :D 

My understanding is that as long as you do not initialize the common debug block (at Fortran side), this block will be considered to be in an "uninitialized state" which stands for "no debug state" (actually, I believe this happens already in 99.99% of cases where you don't need/care about debug traces). Look at the icb examples : arpack is ran twice, the 1st time is performed the classical way (no debug), then you set debug flags (`debug_c`), then you run arpack a 2d time and get debug output (according to flags).

@davydden : does this sounds good to you ? I could also add an icb method like `init_arpack_c`, but, as I can't call that implicitly for arpack users (where to do this ?), any user will have to call it (always) explicitly : no benefit when compared to the "calling `debug_c` approach". Also, the common debug block could be initialized to zero in Fortran, but, I have to say I have no idea how / where to do that !...

Franck

```
~> icb_arpack_c
992.000000
993.000000
994.000000
995.000000
996.000000
997.000000
998.000000
999.000000
1000.000000
------

 _getv0: B-norm of initial / restarted starting vector
 -----------------------------------------------------
    1 -    1:    2.09029D+04
  

 _getv0: B-norm of initial / restarted starting vector
 -----------------------------------------------------
    1 -    1:    2.07939D+04
  
1000.000000 1000.000000
999.000000 999.000000
998.000000 998.000000
997.000000 997.000000
996.000000 996.000000
995.000000 995.000000
994.000000 994.000000
993.000000 993.000000
992.000000 992.000000



```

fghoussen:
By the way

>Consequently, I am quite certain the issue **is** there, but for whatever reason it does not appear on all platforms/compilers/os.

I believe this is a direct consequence of lack of Fortran specifications (<=> "uninitialized state" of the common block) : this probably could be OS/compiler/implementation dependent... No great, but, this is Fortran world. I would say the "preferred" solution  here would be to use `-finit-real=zero -finit-integer=0` at compile time : @davydden , did you try ?

fghoussen:
youpi ....

fghoussen:
@davydden : Seems the last commit (64f3c1b) should fix your problem on your HPC cluster whatever compiler options... Not sure. Does not break anything apparently at my side.... And this is supposed to initialize the debug common block at fortran side : check it and you know ! (I'am not sure - can't test it)

davydden:
> My understanding is that as long as you do not initialize the common debug block (at Fortran side), this block will be considered to be in an "uninitialized state" which stands for "no debug state"

that's not what folks here https://stackoverflow.com/a/49618639/888478 say. I think it's exactly the problem -- this is somehow os/compiler/platform dependent. I have no output on macOS or Ubuntu, but things have broken on HPC cluster.

> -finit-real=zero -finit-integer=0 at compile time

I have not because folks at stackoverflow https://stackoverflow.com/a/49618639/888478 said those flags do not influence COMMON block variables ;-(

> Seems the last commit (64f3c1b) should fix your problem on your HPC cluster

I certainly will! Thanks for not giving up on this issue. 👍

fghoussen:
Seems initialization of debug common block can not be moved to `debug.h` (compil KO) : I reverted commit 22d1b3b !... This means initialization of this common block will be done **only** using icb at configure time...

@sylvestre : if you like this, you may fixup all this misery in one relevant commit !... :D

debug_icb.f90
```diff
@@ -37,3 +37,16 @@ subroutine debug_c(logfil_c, ndigit_c, mgetv0_c,
   mcgets = mcgets_c
   mceupd = mceupd_c
 end subroutine debug_c
+
+! Initialisation of the debug common block to "no debug".
+block data init
```
davydden: can you move this into a separate new `.f90`? It should not be related to ICB and should happen regardless of users having ICB on or off. 

Currently, this source file is not part of the lib unless ICB is on.

debug_icb.f90
```diff
@@ -0,0 +1,52 @@
...
+  mcaup2 = mcaup2_c
+  mcaitr = mcaitr_c
+  mceigh = mceigh_c
+  mcapps = mcapps_c
+  mcgets = mcgets_c
+  mceupd = mceupd_c
+end subroutine debug_c
+
+! Initialisation of the debug common block to "no debug".
+block data init
```
davydden: i would even move the next 10 lines into a stand-alone PR in a separate file like `init_common.f90`. Then if it works (i will check), the changes are un-controversial and i think @sylvestre can merge it quickly.

sylvestre: sure, just ping me when it is ready!

fghoussen:
Pushing from McDonald highway !... If KO, will get back to it next week. If OK, for me, good to go.

sylvestre:
Travis is happy, improvement of the code coverage, taking it.
We can always land follow up commits later.
bravo & thanks

fghoussen:
@davydden : test it on your HPC cluster if you have some time !... This is a blind fix :D

davydden:
> test it on your HPC cluster if you have some time !... This is a blind fix :D

i gave it a try, unfortunately i still see the issues that Arpack tries to print some debug info ;-(

fghoussen:
>i gave it a try, unfortunately i still see the issues that Arpack tries to print some debug info ;-(

Did you `make distclean` or `rm -fr arpack-ng` ?... Before re-bootstrapping + configuring... If yes, then you are welcome to PR if you find the bug !

davydden:
> Did you make distclean or rm -fr arpack-ng ?... Before re-bootstrapping + configuring...

sure, I always build in a clean environment from scratch.

fghoussen:
Too bad ! Blind fixing is definitively not the best way to go... :D

fghoussen:
@davydden : seems you need to "put" the data block in each subroutine or program !?... The more I deal with Fortran, the more I feel bad !

See here http://www.personal.psu.edu/jhm/f90/questions/common.html (how to use block data) and here http://www.personal.psu.edu/jhm/f90/examples/ode/fall.f

So if you feel courageous enough you can "put" that in all subroutines (*[ae]upd*), or, you find THE one initialization method that parpack/arpack may use in all cases (if possible ?). For now my understanding is that the "data block" is here but not called/used !

Hope this helps !...

caliarim:
No, it is not necessary to put the data block in each subroutine. If you put the attached test.f in PARPACK/MPI/SRC and compile with gfortran test.f ./.libs/libarpack.a and run ./a.out you see that the data initialization works both in the program and in the subroutine. I think there is no problem at fortran level.
[test.f.gz](https://github.com/opencollab/arpack-ng/files/1893103/test.f.gz)

davydden:
@caliarim 

> I think there is no problem at fortran level.

I think the problem is there, but I don't know why. What I know and checked, though, is that if I comment out the usage of those common-block debug variables by simply changing
```
msglvl = mgetv0
```
to
```
msglvl = 0
```
then when calling pArpack function from C++ there is no debug output (as epxected). If I don't, even with this PR merged, pArpack tries to print output and fails with 

```
At line 68 of file <blabla>/arpack-ng-3.4.0/PARPACK/UTIL/MPI/pdvout.f
Fortran runtime error: Bad unit number in statement
```

Why exactly it fails is irrelevant, the issue is that it calls printing routines which means those variables are not set to zero despite https://github.com/opencollab/arpack-ng/pull/107/files#diff-ffb8d149270531d336dde96ac0c0704eR11.

caliarim:
The external statement is recommended in order not to forget to add the block data statement. But it is not mandatory. Did you try my test.f?

fghoussen:
Connected with not convenient tablet !... Got an idea. Will test it and let you know tomorrow

fghoussen:
@davydden  : try to add an empty dummy fortran program in debug_init.f90. Does it fix the hpc cluster problem ? Arpack is a lib (probably no main, just subroutines). Data block may be "loaded" before main (= fortran program) to init global variables... No main, no init ?!...

fghoussen:
If KO, try to add " external debug_init " in the dummy prg

fghoussen:
By the way, at my side, I tried: 
```
 >> git diff
diff --git a/debug_init.f90 b/debug_init.f90
index f4c4f8e..74ae137 100644
--- a/debug_init.f90
+++ b/debug_init.f90
@@ -10,3 +10,7 @@ block data debug_init
        mcaupd, mcaup2, mcaitr, mceigh, mcapps, mcgets, mceupd &
        / 6, -3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 /
 end block data debug_init
+
+program dummy
+  external debug_init
+end program dummy

```
This does not seem to break anything at my side but I don't know if it can fix the cluster problem !

fghoussen:
```
+program dummy
+  include 'debug.h'
+  external debug_init
+end program dummy

```

Seems to be OK too !...

@caliarim : I've just looked at your test.f, I would say (not sure) it works _because_ there is a `program` !

PR #109, opened 2018-04-09
iso_c_binding: add statistics support.

fghoussen:
Should be the last PR !... :D Hope coverage will not break

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,9 @@ arpack-ng is replacing arpack almost everywhere.
 
    The following instructions explain how to make the ARPACK library.
 
-4. Unlike ARPACK, ARPACK-NG is providing autotools based build system.
+4. Unlike ARPACK, ARPACK-NG is providing autotools based build system
+   and iso_c_binding support (which enables to call fortran subroutines
+   natively from C or C++).
 
 Therefore, the classical commands should work as expected:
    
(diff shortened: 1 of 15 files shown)


sylvestre:
We can regress the code coverage if we need to.

fghoussen:
icb is "just" typing everything (function, suroutine, arguments, ...) : it's tedious + lot of declarations to (re-)do : so coverage may logically drop ! :D We'll see...

sylvestre:
Well done, no regression in term of perf.
Shouldn't this change be mentioned in the changelog?

fghoussen:
Sure ! I've added 2 lines in changelog and pushed force

README.md
```diff
@@ -70,7 +70,9 @@ arpack-ng is replacing arpack almost everywhere.
 
    The following instructions explain how to make the ARPACK library.
 
-4. Unlike ARPACK, ARPACK-NG is providing autotools based build system.
+4. Unlike ARPACK, ARPACK-NG is providing autotools based build system
+   and iso_c_binding support (which enables to call fortran subroutines
```
sylvestre: maybe you could add a link to a page which explains what is iso_c_binding?

fghoussen: Woaw !... Not so easy. I try to get one.
I know a good one in french (in idris website - where I learnt everything from scratch), but, if I remember well this one is not translated in english !... I'll try to find something.

fghoussen: My understanding is that IDRIS lectures are french only. But I found:
http://fortranwiki.org/fortran/show/ISO_C_BINDING
http://fortranwiki.org/fortran/show/Generating+C+Interfaces

Do you buy one of these or both ?!

sylvestre: Yeah, clearly!

PR #116, opened 2018-05-20
Add clang-format

dbeurle:
(Optional) add clang-format for consistent formatting


sylvestre:
great idea!
just a comment, I would prefer using a well defined coding style like Google's, could you please change that?

dbeurle:
Absolutely - I put the google style in.  Do you know of a way to get CI to make sure this is correct / do the formatting on commit?  Just in case a dev doesn't have clang-format installed locally.

sylvestre:
`BasedOnStyle: Google`
Just that should be enough, don't you think?

dbeurle:
Something new every day.  I did `clang-format -style=google -dump-config > .clang-format`.  Can you squash these commits on merge?

PR #117, opened 2018-05-20
Add C++11 interface

dbeurle:
Use GNU extensions otherwise _Complex_I is not valid.


dbeurle:
See #86

dbeurle:
@sylvestre I can reformat this based on the clang-format I PR'd before if that gets pulled.

sylvestre:
Please mention this change in the release notes!
Thanks again

PR #119, opened 2018-05-21
Remove appended _c and update comments

dbeurle:
Per review.  Formatting changes to match Google format.


dbeurle:
After this I've got everything done you can finally release ^_^

fghoussen:
What about parpack ?

dbeurle:
Is it all working?  The same patterns should apply.

fghoussen:
For symmetry, would be good you do the job for parapck too

sylvestre:
Indeed, @dbeurle could you do that? Thanks :)

dbeurle:
I can - it'll have to be tonight since unfortunately I don't get paid to write code :-(

PR #122, opened 2018-05-29
Add includedir to pkg-config file

ypnos:
Without this line, pkg-config 0.29.2 refuses to parse the arpack.pc file:

# pkg-config --libs arpack
Variable 'includedir' not defined in '/usr/lib/pkgconfig/arpack.pc'


sylvestre:
Thanks!

Issue #123, opened 2018-06-05
icb_arpack_cpp: testsuite fails on i386


sylvestre:
See: https://buildd.debian.org/status/fetch.php?pkg=arpack&arch=i386&ver=3.6.0-1&stamp=1528229606&raw=0

```
../build-aux/test-driver: line 107:  7229 Segmentation fault      "$@" > $log_file 2>&1
FAIL: icb_arpack_cpp
../build-aux/test-driver: line 107:  7200 Segmentation fault      "$@" > $log_file 2>&1
FAIL: icb_arpack_c
```

@dbeurle @fghoussen Rings a bell?

10110111:
For reference, here's the backtrace of the crash and a bit more of info:
```
Thread 1 "lt-icb_arpack_c" received signal SIGSEGV, Segmentation fault.
cneupd (rvec=.TRUE., howmny=..., select=<error reading variable: value requires 538300256 bytes, which is more than max-value-size>, d=<error reading variable: value requires 1076211256 bytes, which is more than max-value-size>, 
    z=<error reading variable: value requires 3545643256 bytes, which is more than max-value-size>, ldz=1001, sigma=(0,0), workev=<error reading variable: value requires 2153201024 bytes, which is more than max-value-size>, bmat=..., n=134526416, 
    which=<error reading variable: Cannot access memory at address 0x3e8>, nev=134526407, tol=1.26116862e-44, resid=<error reading variable: value requires 1076211328 bytes, which is more than max-value-size>, ncv=134575064, 
    v=<error reading variable: value requires 1102378496 bytes, which is more than max-value-size>, ldv=-134864888, iparam=<error reading variable: Cannot access memory at address 0x3e8>, ipntr=..., workd=<error reading variable: value requires 3228633984 bytes, which is more than max-value-size>, 
    workl=<error reading variable: value requires 1076856640 bytes, which is more than max-value-size>, lworkl=134607080, rwork=<error reading variable: value requires 538300256 bytes, which is more than max-value-size>, info=0, _howmny=1, _bmat=1, _which=1) at cneupd.f:340
340           mode = iparam(7)
(gdb) p iparam
Cannot access memory at address 0x3e8
(gdb) bt
#0  cneupd (rvec=.TRUE., howmny=..., select=<error reading variable: value requires 538300256 bytes, which is more than max-value-size>, d=<error reading variable: value requires 1076211256 bytes, which is more than max-value-size>,                                                                                    
[...]

PR #124, opened 2018-06-06
Pass complex parameters by pointers to constant instead of by value

10110111:
Fortran-C bindings seem to expect it.

This fixes opencollab/arpack-ng#123. Both C and C++ tests will no longer
crash. But the C test will fail (but still pass on x86_64). This might be due to an error in the
test itself.


dbeurle:
The complex C test appears to be the same as for the c++ one.  I can write a test for the `double` case in C++ to rule that out, if that would help?

10110111:
Actually a good idea. I've simply replaced `s/\<float\>/double/g` in the relevant pieces of C++ test, and now it crashes. I suppose I'll retract this pull request for now.

PR #131, opened 2018-06-23
Fix issue#130 for cmake.

### LINKED ISSUE #130: Bad header file names
The C bindings install
```
   /usr/include/arpack.h
   /usr/include/arpack.hpp
   /usr/include/debug-arpack.h
   /usr/include/debug_c.h
   /usr/include/debug_c.hpp
   /usr/include/stat.h
   /usr/include/stat_c.h
   /usr/include/stat_c.hpp
```
The first three are fine, but the last 5 aren't okay since the names are way too generic.

These files should either be installed in an arpack specific directory, or prefixed with ```arpack_```.

fghoussen:
Should fix #130 for cmake. Got no idea how to fix this with autotools !...


sylvestre:
could you please mention that in the changelog?

fghoussen:
autotools: hope make distcheck will be OK !?...

fghoussen:
kill 678f6ce if you don't want it (?)

PR #133, opened 2018-06-25
C++11 (for ICB): do NOT use hard-coded gnu++11.

fghoussen:
Hard-coded gnu++11 may break on specific archis (armel, ppc64el, ...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,13 @@ addons:
       - gfortran
       - gcc
       - g++
+      - openmpi-bin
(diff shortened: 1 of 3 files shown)


fghoussen:
Not sure you want to merge that...
Not sure travis will be OK: wait and see...

dbeurle:
You'll have to also disable it explicitly in CMake as by default you get compiler extensions.  The problem is the use of complex which is a GNU only extension.

Issue #140, opened 2018-07-05
Increase SONAME version numbering in line with libtool suggestions


renierm:
The SONAME version was last updated with the 3.0 release of arpack-ng.

The SONAME version numbering should be updated for code revisions as well (revision number only still indicating interface compatibility). If this is not updated it makes it difficult to track that a library is in fact newer.

See ftp://ftp.gnu.org/old-gnu/Manuals/libtool-1.4.2/html_node/libtool_35.html

sylvestre:
The soname is about the versioning of the ABI, not the software itself.
AFAIK, the arpack library is still ABI compatible with the previous versions. Therefor, doesn't require a bump of soname. 

In the Debian package, I am maintaining a list of symbols https://salsa.debian.org/science-team/arpack/blob/master/debian/libarpack2.symbols and the tool didn't detect changes which would requires to bump the SONAME.

renierm:
Regardless if it has to do with API or ABI it is still helpful to change the revision number if (quoting from the link above)

> 3. If the library source code has changed at all since the last update, then increment revision (`c:r:a' becomes `c:r+1:a')

The problem that I am facing is that if I compile with arpack-ng-3.4.0 or arpack-ng-3.6.1 I end up with libarpack-ng.so.2.0.0. So there is no way of telling which library I am actually dealing with when looking at the file itself.

If the revision number is updated as suggested then I would at least have say libarpack-ng.so.2.0.1. and libarpack-ng.so.2.0.4. for argument sake. Both would be compatible with my binary and I would use soft links (as is customary - libarpack-ng.so.2) to point to the desired version.

renierm:
Thanks.

I have submitted a pull request for this issue, however this is only for this release and I will not be able to do this for every release. I would suggest that you add this to the stuff that needs to be done for every release.

I don't think it would make sense to go an try to fix this for versions in the past as that might break many things for people relying on the full SONAME that currently assume 2.0.0.

ps. this is my first pull request so pardon if I missed something.

PR #141, opened 2018-07-05
Update SONAME to reflect the change in source code

renierm:
Update the SONAME for the latest version as discussed in https://github.com/opencollab/arpack-ng/issues/140.

### DIFF
--- a/PARPACK/SRC/BLACS/Makefile.am
+++ b/PARPACK/SRC/BLACS/Makefile.am
@@ -20,4 +20,4 @@ libparpack@LIBSUFFIX@_la_LIBADD = \
 	$(top_builddir)/PARPACK/UTIL/BLACS/libparpackutil.la \
 	$(top_builddir)/SRC/libarpack@LIBSUFFIX@.la \
 	$(LAPACK_LIBS) $(BLAS_LIBS)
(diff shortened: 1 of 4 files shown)


sylvestre:
First, many thanks for that

Two comments:
* I think we should move that into the configure.ac to have a single declaration
* also update the cmake build

If you don't know how to do that, please let me know :)

renierm:
I will get back to this sometime this week, just a bit swamped with work :-)

renierm:
I have pushed the change to only define SONAME once in configure.ac which is then propagated to the relevant Makefile.am files. The value is currently directly set in configure.ac which is maybe not be best place as this could be read from an environment variable?

I am not quite sure what needs to happen with regards to the cmake build.

sylvestre:
I didn't forget about you. We (or maybe I) need to fix the cmake stuff before landing this patch.

sylvestre:
Closing until someone propose a cmake fix too.

fghoussen:
Not an expert here. Looked quickly to https://cmake.org/cmake/help/v3.13/prop_tgt/SOVERSION.html.
Seems this is "just" about adding :
set(SOVERSION "2:1")
in CMakeLists.txt ? Or something similar ?

Issue #144, opened 2018-08-21
Tests failing with: "Fortran runtime error: Unit number in I/O statement too small"


markuskowa:
When I try to build `arpack-ng 6.3.2`, some of the tests fail with
`Fortran runtime error: Unit number in I/O statement too small`

Looking at the code it seems that the variable `logfil` is not initialized properly for the failing tests.

### Details ###

- arpack 6.3.2 (problem first occurs in 3.6.0)
- NixOS unstable
- compiler: gfortran 7.3
```
PASS: dnsimp
PASS: bug_1323
FAIL: bug_58_double
PASS: bug_79_double_complex
FAIL: bug_1315_single
FAIL: bug_1315_double
```
[...]
```
At line 41 of file ivout.f
Fortran runtime error: Unit number in I/O statement too small

Error termination. Backtrace:
#0  0x7ffff5a5066d in ???
#1  0x7ffff5a51175 in ???
#2  0x7ffff5a518c9 in ???
#3  0x7ffff7bba6e6 in ivout_
        at /build/arpack-ng-3.6.2/UTIL/ivout.f:41
#4  0x7ffff7b9b632 in dnaitr_
        at /build/arpack-ng-3.6.2/SRC/dnaitr.f:375
#5  0x7ffff7b9e221 in dnaup2_
        at /build/arpack-ng-3.6.2/SRC/dnaup2.f:362
#6  0x7ffff7b9f190 in dnaupd_
        at /build/arpack-ng-3.6.2/SRC/dnaupd.f:606
#7  0x401275 in dndrv2
        at /build/arpack-ng-3.6.2/TESTS/bug_58_double.f:232
#8  0x400dbe in main
        at /build/arpack-ng-3.6.2/TESTS/bug_58_double.f:430
FAIL bug_58_double (exit status: 2)
```
(see https://gist.github.com/markuskowa/6d13a14f18fa6340f7d55cfb09998d2b for full output)

fghoussen:
What are your arch ? OS ? compiler ? compiler version ? Did you try with 2 different compilers (say gcc/intel) ? Possible duplicate of #82 ? Does the pb disappear when using `--enable-icb` ?

fghoussen:
bug_58_double looks full fortran: probably not fixed by icb !
When using gcc, try playing with `-f-init-*` options (`export FCFLAGS` before `configure`) : this may at least bypass the issue (if correction could not be found/done)

fghoussen:
[...]

PR #146, opened 2018-08-22
Fix initialization of debug variables

markuskowa:
Replaces `debug_init.f90` with `debug_init.f`.
Fixes https://github.com/opencollab/arpack-ng/issues/144

MPI version is untested.


markuskowa:
Travis still shows pending but when I click on on it say that all tests have passed. Did I miss something here.

sylvestre:
I restarted the build

markuskowa:
Thanks! looks good now.

fghoussen:
Probably not the "good" correction

PR #152, opened 2018-09-15
[CLEAN] Run MPI tests with "mpirun -n 2".

fghoussen:


fghoussen:
KO: seems to be fake timeout job !... @sylvestre : could you re run it ?

sylvestre:
done!

PR #153, opened 2018-09-15
[BUG FIX] howmny is defined as character*1 in pdseupd subroutine.

fghoussen:
Cosmetic bug ("safer" way to get same thing).


fghoussen:
KO: seems to be fake timeout job !... @sylvestre : could you re run it ?

sylvestre:
Retriggered!

Issue #149, opened 2018-09-07
not able to compute eigenvalues for generalized non-hermitian eigenvalue problem. 


manavbhatia:
Hi,

   I am having trouble computing eigenvalues with arpack for the following eigenvalue problem: 

```
   M1 x = lambda M0 x 
```

   The matrices M1 and M0 are 178x178 and are attached with this email. They are from a finite element discretization of a beam. 

   While LAPACK’s dggev is able to compute the eigenvalues for my matrices without issues, I am unable to get arpack to do this form me. I am using “dnaupd" and it is returning with "info=-8". I am also getting the same error when attempting to use "eigs" command in octave. 

    I have struggled with this for a few days and am not sure how to proceed. I have made sure that the driver routine calling “dnaupd" is fine, and the same error from octave gives me some assurance on the reproducibility. 

    I would greatly appreciate if you could provide some guidance on how to identify the issue here. 

[M0.txt](https://github.com/opencollab/arpack-ng/files/2359509/M0.txt)
[M1.txt](https://github.com/opencollab/arpack-ng/files/2359510/M1.txt)

caliarim:
Which eigenvalues are you trying to compute? Which mode?

manavbhatia:
I have tried both smallest_magnitude and largest_magnitude with no success. I have played around with number of modes from 2 to 20.

caliarim:
I'm away from my computer. Is it possible that M0 is not SPD? Can you compute chol (M0) in octave?

manavbhatia:
Yes, I am able to compute chol(m0) and chol(m1).

caliarim:
Ok, I will give a look when back from my holidays.

manavbhatia:
Here is a related thread on octave concerning the same issue: http://octave.1599824.n4.nabble.com/Issues-with-eigs-td4689314.html#a4689320 .

While different people are able to converge to somewhat different eigenvalues on different releases of octave, I did not get convergence using octave 4.4.1 on Mac OS with arpack installed through macports. I am linking to the same arpack in my user code.

caliarim:
Which version of arpack are you using?

manavbhatia:
v 3.6.2 (installed from macports on Mac 10.13.6).

fghoussen:
@manavbhatia : did you try with slepc ?
[...]

PR #155, opened 2018-09-15
[BUG FIX] ICB: missing workev for *[ds]neupd (real+not-sym).

fghoussen:
Even with real matrices, eigen values can be complex if matrices
are not symmetric. In this case, you need workev.

ICB tests: should be OK as it turns out that they do NOT use
real+not-sym methods (they use real+sym and complex+non-sym).


sylvestre:
You are changing the ABI. 
I don't think I can take this patch this way :/

fghoussen:
You're right ! Missed that... But, as is, there _is_ a problem (if I am not mistaken) : the current API is screwed for *[ds]neupd (likely nobody tried, but, using *[ds]neup should just not compile !... Or crash if you skip `workev`). Time for new release ?...

If people around could double check this... Would be a good thing (I do believe this is needed but as there are many many possible combinaisons, not so easy to clear mind)

fghoussen:
Or I could try to add up the utility (#149) to this PR... Waiting for new release. Let me know !

fghoussen:
OK: branch rebased on master + add line in changelog (change API) + pushed force back here !
Busy theses days: if I can save time, I'll try to finish backporting utility (old -> current code base), and, PR it on top of that PR.

sylvestre:
I don't think we want to change the ABI/API here. arpack has been stable for a few decades. Would be a pain for everyone.
sorry :/

fghoussen:
This is **only** changing ABI/API of the ICB wrappers (*[ds]neupd_**c** - ICB adds these symbols along with native fortran symbols *[ds]neupd_) : the reference ABI/API is **not** changed (native fortran symbols like *[ds]neupd_ and / or *[ds]neupd__ will **not** be modified).

When using ICB,  you'll use *[ds]neupd_**c** (modified) who will trigger a call to *[ds]neupd_ (**not** modified).

There is not much choice. As is, if you include "arpack.h" (= using ICB) and if you use *[ds]neupd_**c**, you get:

- a crash if you forgot about the extra (missing) `workev` argument at C side (assuming the fortran compiler doesn't check anything and compile anyway !...)

- a compilation error if you remember about `workev` and add  it at C side (call would not be consistent with header definitions)

```
>> ./configure --prefix=/tmp/local; make all install
>> nm /tmp/local/lib/libarpack.so | grep dseupd
000000000001e100 T dseupd_
0000000000022820 T dseupd_c
```

sylvestre:
oh, ok, icb is new enough then :)
Could you rebase the changelog as I released a new version of arpack?

fghoussen:
Done ! Busy these days, no time to finish utility : when do you plan to release a new version ?

sylvestre:
you forgot to update the changelog :)

fghoussen:
I added a line in the CHANGES file. You want me to update the README.md ? The doc ?

sylvestre:
3.6.3 has been released. https://github.com/opencollab/arpack-ng/releases/tag/3.6.3
trunk = 3.6.4 now

fghoussen:
rebased ! (#157  needs this)

fghoussen:
Re-rebased !... :D

fghoussen:
The last job has stalled. Needs to be re-run

fghoussen:
So now, let's handle eigen (which turns out to CD in autotools error traces) => try simple first: apt-get upgrade... New commit is coming...

asartori86:
@sylvestre, would this patch be merged soon? thanks

PR #157, opened 2018-09-23
Arpackmm

fghoussen:
Do NOT merge this PR ! :D This is likely **not** yet working.

If this could go to a happy end, be aware that this PR is based on #155 : merging this PR implies merging #155.

Just want to share the idea (utility in connection with #149) , see if CI is OK (find eigen, test will be KO)... And get some help !

Currently: EXAMPLES/MATRIX_MARKET/arpackmm.sh fails for reason I can't figure out. 
@caliarim , @manavbhatia, anyone else: any help is welcome !...

To test this : 

```
~> git clone https://github.com/fghoussen/arpack-ng
~> git checkout arpackmm
~> sudo apt-get install libeigen3-dev
~> ./bootstrap; ./configure --enable-icb-exmm; make all check
~> tail -n 100 EXAMPLES/MATRIX_MARKET/test-suite.log
```

Problem: computed eigen vector is just wrong ! This is likely because the code "doesn't know" how to use the cubersome arpack lib !...

PR #164, opened 2018-11-16
TravisCI - move CI to xenial: get not-so outdated packages.

fghoussen:
I believe it would be a good thing to move CI to xenial to get "more updated" packages.
autotools jobs break: circular dependency. Don't know how to fix this. If somebody can help...


fghoussen:
This PR is related to #157

sylvestre:
Could you please see if we could run travis on the current Ubuntu + xenial?
This because arpack is used on old systems and I would like to keep some CI on "old" systems.
Thanks

fghoussen:
> Could you please see if we could run travis on the current Ubuntu + xenial?

That's what this PR is meant to (try to see if it's possible or not)

> This because arpack is used on old systems and I would like to keep some CI on "old" systems.

You mean 2 CIs ? one for ubunty trusty + one for unbuntu xenail if possible (with matrix in yml ?)

sylvestre:
correct

fghoussen:
OK, I can first try to test "stuffs" to see if this makes ubuntu + xenial OK.

If xenial turns out to be OK, then it would be possible to add matrix in yml to get CI on both ubuntu xenial and trusty (or precise ?)

fghoussen:
Still failing ?!...

```

make[2]: Circular ../../debug_c.h <- ../../debug_c.h dependency dropped.
cd . && /bin/bash ./config.status ../../debug_c.h
config.status: error: invalid argument: `../../debug_c.h'
Makefile:468: recipe for target '../../debug_c.h' failed

```

Don't get why ?!...

fghoussen:
Who Wants to Be a Millionaire ?
Q: How to fix these circular dependencies ?
A: Asking for a friend...

fghoussen:
@sylvestre: note that e3ce42e does not seem to be _exactly_ a "make distcheck". My understanding is that it makes a tarball but doesn't re-deploy/test it (like distcheck does as far as I understand)

fghoussen:
Giving up: no solution for now...

fghoussen:
@sylvestre : ignoring the crap, sounds good to you ? want to keep e3ce42e ?

fghoussen:
still need to include matrix in yml for precise. not sure how to do that

fghoussen:
CMakeLists.txt is supported on not-so-outdated xenial, but *not* on outdated trusty... great...

fghoussen:
@sylvestre : you may allow concurrent jobs in travis ci setting at opencollab side, would get faster ci !

fghoussen:
fixed up to 2 commits : distcheck fix + travis modif

.travis.yml
```diff
@@ -23,15 +23,55 @@ addons:
...
+    - stage: precise-autotools-seq
+      dist: precise
+      script: ./bootstrap && ./configure                                                                && make VERBOSE=1 && make check && make distcheck;
+    - stage: precise-autotools-mpi
+      dist: precise
+      script: ./bootstrap && ./configure --enable-mpi                                                   && make VERBOSE=1 && make check && make distcheck;
+    # trusty <=> test "older" systems, without ICB
+    - stage: trusty-cmake-seq
+      dist: trusty
+      script: mkdir -p build && cd build && cmake -D EXAMPLES=ON -D MPI=OFF -D ICB=OFF               .. && make VERBOSE=1 && make test  && make package_source;
```
fghoussen: CMake "make package_source" <=> autotools "make dist" only, **not** "make dist**check**". You may drop that ?

fghoussen:
@sylvestre : note that circular dependencies seem not to be solved yet (still appear in traces). But at least they no more break the CI which could have been extended to precise, trusty and xenial.

fghoussen:
Should be all good. Need to merge this before #157

PR #175, opened 2019-01-10
Add back the coveralls push

sylvestre:
bring back the code coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ jobs:
   # xenial <=> coverage: "recent" systems with ICB
   - stage: coverage
     dist: xenial
-    script: travis_wait mkdir -p build && cd build && cmake -DEXAMPLES=ON -DMPI=ON -DICBEXMM=ON -DCOVERALLS=ON .. && make && make test coveralls;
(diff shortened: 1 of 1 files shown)


fghoussen:
Oh ?! Didn't noticed that. No, don't know what's going on ?!... Would say this is broken since .travis.yml has been rewritten (by me !...) with `stages` instead of `script` ?

fghoussen:
Got it ! Look at the logs
> The job exceeded the maximum log length, and has been terminated.

You need to use travis_wait : https://docs.travis-ci.com/user/common-build-problems/

fghoussen:
After thinking of it, it may not work. If so, revert `travis_wait` but try `make all test > /dev/null  ` : the message says log length limit is reached (not the time limit - I thought at first it was that)

fghoussen:
worked ! :D still weird : it was working before ?!...

sylvestre:
I guess the log was shorter?

fghoussen:
yeah... maybe the log was previously touching the limit.
arpackmm is likely to have increased coverage : is so, you may see the coverage rate increased.

PR #176, opened 2019-01-10
Make coverage the first job to execute

sylvestre:
After info here as coverall will run in parallel


sylvestre:
Looks like the order doesn't matter

fghoussen:
No order doesn't matter: stages are independents.

PR #181, opened 2019-01-10
Arpackmm: add --schur option.

fghoussen:
## Pull request purpose

arpackmm: enhancement.

Enable to compute ritz or schur vectors.

### DIFF
--- a/EXAMPLES/MATRIX_MARKET/arpackmm.cpp
+++ b/EXAMPLES/MATRIX_MARKET/arpackmm.cpp
@@ -59,6 +59,7 @@ class options {
       invert = false; // Eigen value invertion: look for 1./lambda instead of lambda.
       tol = 1.e-06;
       maxIt = 100;
+      schur = false; // Compute Ritz vectors.
       slv = "BiCG";
       slvItrTol = nullptr;
       slvItrMaxIt = nullptr;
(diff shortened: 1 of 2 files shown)


sylvestre:
Is it possible to have something less cryptic ? slvDrtPvtThd

fghoussen:
This is based on #180 + coverage is likely to increase.
That's all I got for arpack-3.7 !

fghoussen:
> Is it possible to have something less cryptic ? slvDrtPvtThd

What do you propose ?

Lots of options already around ! To know what they deal with prefix are added up

```
~> ./arpackmm --help
Usage: running arpack to check for eigen values/vectors.
...
  --slv S:          solver (BiCG, CG, LU)
                      BiCG: iterative method, any matrices
                        CG: iterative method, sym matrices only
                        LU:    direct method, any matrices
                        QR:    direct method, any matrices
                    default: BiCG
  --slvItrTol T:    solver tolerance T (for iterative solvers).
                    default: eigen default value
  --slvItrMaxIt M:  solver maximum iterations M (for iterative solvers).
                    default: eigen default value
  --slvDrtPvtThd T: solver pivot threshold T (for direct solvers).
                    default: eigen default value

```

pivot threshold only apply to direct (Drt) solver.
tol / max it only apply to iterative solver (Itr).

Probably more options to go when in version > 3.7, I'll had to add preconditioner (which make sense only with iterative solver : so you probably could expet a --slvItrPC)

fghoussen:
rebased. no coverage increase: too bad !

Issue #183, opened 2019-01-12
error message could be improved when eigen isn't available


sylvestre:
./configure: line 25605: syntax error near unexpected token `EIGEN3,'
./configure: line 25605: `	PKG_CHECK_MODULES(EIGEN3, eigen3 >= 3.2)'
	tail -v -n \+0 config.log
==> config.log <==

fghoussen:
I have the same kind of errors when a package is not found. 

> ~>pkg-config --version
> 0.29

wshuai294:
could I ask how to solve the problem? I met the same error 
```
./configure: line 26353: syntax error near unexpected token `EIGEN3,'
./configure: line 26353: `      PKG_CHECK_MODULES(EIGEN3, eigen3 >= 3.3)'
```
and I have installed `pkg-config 0.29.2`, it did not solve the problem.

fghoussen:
If you bootstrapped before installing `pkg-config 0.29.2`, try

```
>> make distclean
>> ./bootstrap
>> ./configure
```

MacroUniverse:
I also got an error for the latest version 3.8.0 on ubuntu 22.04
My command is just `sh bootstrap`, then `./configure`. As provided in README
```
./configure: line 26616: syntax error near unexpected token `EIGEN3,'
./configure: line 26616: `	PKG_CHECK_MODULES(EIGEN3, eigen3 >= 3.3)'
```
Does arpack-ng require Eigen? If so, when I run `./configure --help`, there is no option for providing the path to Eigen, what should I do to set the path?

My temporary solution is to simply delete the line 26616 of `./configure`  file. And everything seems to be fine, `make check` passed all tests.

fghoussen:
>Does arpack-ng require Eigen? 

No, `arpack-ng` needs `eigen` only when asking for `exmm` option (matrix market examples using C/C++ bindings allowing direct use of fortran).

Can you print with `AC_MSG_RESULT` what `$enable_icb_exmm` contains (before `if test`) and report here? (I do not reproduce the problem)
https://github.com/opencollab/arpack-ng/blob/db55a7ff4180bb440a980d618e302eded0db52f3/configure.ac#L92
https://github.com/opencollab/arpack-ng/blob/db55a7ff4180bb440a980d618e302eded0db52f3/configure.ac#L93

MacroUniverse:
I added a line `AC_MSG_RESULT($enable_icb_exmm)`, then ran `./configure` again. It shows,
```
[...]

PR #185, opened 2019-01-12
Add error message if eigen is not found at configure time (autotools).

fghoussen:
## Pull request purpose

fixing issue #183 ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,7 +77,7 @@ AC_ARG_ENABLE([icb-exmm],
               [AS_VAR_SET([enable_icb_exmm], [no])])
 
 if test x"$enable_icb_exmm" != x"no"; then
-	PKG_CHECK_MODULES([EIGEN3], [eigen3 >= 3.2])
+	PKG_CHECK_MODULES([EIGEN3], [eigen3 >= 3.2], [], [AC_MSG_ERROR([Eigen not found. Give hints with PKG_CONFIG_PATH])])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -77,7 +77,7 @@ AC_ARG_ENABLE([icb-exmm],
               [AS_VAR_SET([enable_icb_exmm], [no])])
 
 if test x"$enable_icb_exmm" != x"no"; then
-	PKG_CHECK_MODULES([EIGEN3], [eigen3 >= 3.2])
+	PKG_CHECK_MODULES([EIGEN3], [eigen3 >= 3.2], [], [AC_MSG_ERROR([Eigen not found. Give hints with PKG_CONFIG_PATH])])
```
sylvestre:
```suggestion
	PKG_CHECK_MODULES([EIGEN3], [eigen3 >= 3.2], [], [AC_MSG_ERROR([Eigen not found. Give hints with PKG_CONFIG_PATH. Disable the need of this dependency using --disable-icb-exmm])])
```

sylvestre: what do you think?

fghoussen: sure !

fghoussen:
a change in configure.ac changes coverage ?... really ?...

fghoussen:
Error added in configure (by configure.ac) is NOT covered (would break CI) : is this where the reg comes from ?!... :D

PR #186, opened 2019-01-13
Autoconf: add enable-atlas/openblas.

fghoussen:
## Pull request purpose

autotools: enhancement 

- add --enable-atlas
- add --enable-openblas

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ stages:
   - name: precise
   - name: trusty
   - name: xenial
+  - name: openblas
+  - name: atlas
   - name: coverage
   - name: interface64
 
@@ -86,6 +88,14 @@ jobs:
   - stage: xenial
     dist: xenial
     script: ./bootstrap && ./configure --enable-mpi --enable-icb                        && make VERBOSE=1 && make check && make distcheck;
+  # xenial: test openblas support with autoconf
+  - stage: openblas
+    dist: xenial
+    script: sudo apt-get -y install libopenblas-dev && ./bootstrap && ./configure --enable-openblas --enable-mpi --enable-icb && make VERBOSE=1 && make check && make distcheck;
+  # xenial: test atlas support with autoconf
+  - stage: atlas
+    dist: xenial
+    script: sudo apt-get -y install libatlas-base-dev && ./bootstrap && ./configure --enable-atlas --enable-mpi --enable-icb && make VERBOSE=1 && make check && make distcheck;
   # xenial <=> coverage: "recent" systems with ICB
   - stage: coverage
     dist: xenial
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 
 [ Franck Houssen ]
  * arpackmm: add --slvItrPC option (PC: Jacobi, ILU).
+ * Add --enable-atlas/openblas in configure script for convenience.
 
 arpack-ng - 3.7.0
 
--- a/PARPACK/TESTS/MPI/icb_parpack_cpp.cpp
+++ b/PARPACK/TESTS/MPI/icb_parpack_cpp.cpp
@@ -86,8 +86,8 @@ void real_symmetric_runner() {
                 workl.data(), lworkl, info);
 
   for (int i = 0; i < nev; ++i) {
-    std::cout << "rank " << rank << " - " << d[i] << std::endl;
-    if (std::abs(d[i] - static_cast<float>(1000 - (nev - 1) + i)) > 1e-1) {
+    std::cout << "rank " << rank << " - " << d[i] << " vs " << static_cast<float>(1000 - (nev - 1) + i) << std::endl;
+    if (std::abs(d[i] - static_cast<float>(1000 - (nev - 1) + i)) > 1.) { // Simple precision, relax tol.
       throw std::domain_error("Correct eigenvalues not computed");
     }
   }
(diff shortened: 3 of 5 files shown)


fghoussen:
Not sure you wanna buy that !...

sylvestre:
atlas is failing
and it would be nice to have cmake too
(and CHANGES updated)
i would take that :)

fghoussen:
OK... Quite busy today: will be back on that tonight or tomorrow.
If I am not mistaken, the FindBLAS/LAPACK.cmake is supposed to already handle atlas/openblas/mkl according to what user wants.

fghoussen:
quick look, no idea. @turboencabulator : any idea ?

fghoussen:
@sylvestre: can't access logs of atlas job at my side in travis GUI ?! can you restart the job ?

fghoussen:
logs are back (?!), but, no solution for now

fghoussen:
dead ci ?!

turboencabulator:
ax_blas.m4 handles the library search order, it's preferring openblas, then atlas, then several others.  It also lets you override it and specify which library you want, so you should be able to get this same effect without having to change the configure script. Something like `--with-blas=openblas` or `--with-blas=satlas`.  Not really sure about atlas since I think they've changed their library naming at some point.

With these changes it's going to be looking for particular pkgconfig files which might not even exist.  On my machine I have openblas.pc (not the blas-openblas.pc or lapack-openblas.pc that it's searching for).  I have a package of an earlier version of openblas that doesn't include any pkgconfig file, and my latest atlas package doesn't have one either.  Are these files Debian-isms that other distros won't have?

I'm thinking it would be better to document the appropriate `--with-blas` flags to make it work with your desired library rather than add new flags that may not work in all cases.  I assume CMake probably has some equivalent -D flag to do this?

fghoussen:
> ax_blas.m4 handles the library search order, it's preferring openblas, then atlas, then several others. It also lets you override it and specify which library you want, 

Yes, I know (I use that in interface stage in travis.yml) !
... But I always felt like it's kind of cumbersome (or say not convienient).

> so you should be able to get this same effect without having to change the configure script.

Yes, if you are "expert enough" (means you know how to use pkg-config + how to pass that to ax_blas.m4 - lots of students I know don't ! :D)

> Something like --with-blas=openblas or --with-blas=satlas. Not really sure about atlas since I think they've changed their library naming at some point. With these changes it's going to be looking for particular pkgconfig files which might not even exist. 

No big deal (to me). If file is not found: do it by hand (even if it's not "so convenient")

> On my machine I have openblas.pc (not the blas-openblas.pc or lapack-openblas.pc that it's searching for). I have a package of an earlier version of openblas that doesn't include any pkgconfig file, and my latest atlas package doesn't have one either. Are these files Debian-isms that other distros won't have?

Do you have associated *-dev packages installed ? If yes, and, If pc file names change accross distro, this PR is likely meant to be closed !

> I'm thinking it would be better to document the appropriate --with-blas flags to make it work with your desired library rather than add new flags that may not work in all cases. I assume CMake probably has some equivalent -D flag to do this?

According to the doc (https://cmake.org/cmake/help/v3.13/module/FindBLAS.html), it seems that FindBLAS already handle customizing the search by setting stuffs like `set BLA_VENDOR=OpenBLAS` before running `cmake`... But from my unlucky experience, this is _not_ always working like people would expect it to work !

fghoussen:
what if it's "just" a precision problem as we use atlas + simple precision arpack functions (s[ae]upd) ?

fghoussen:
indeed, looks like a precision problem (after blas/lapack is replaced by atlas' ones).
double check is welcome !...

fghoussen:
> On my machine I have openblas.pc (not the blas-openblas.pc or lapack-openblas.pc that it's searching for). I have a package of an earlier version of openblas that doesn't include any pkgconfig file, and my latest atlas package doesn't have one either

9c77ab2 may be a way to deal with that ?...

personal opinion: always prefer last updated versions + think "ahead" and use new arch/package (not target old-cray-or-archaic arch/package - the day you're stuck here: use ax_blas the "old" way specifying libs by hand...)

fghoussen:
@turboencabulator , @sylvestre : what do you say ? is this worth ?

personal opinion: not "perfect" (what's perfect ?), but, convenient in many cases.

fghoussen:
CI KO ! :D _Really_ not perfect !

turboencabulator:
> Do you have associated *-dev packages installed ? If yes, and, If pc file names change accross distro, this PR is likely meant to be closed !

I'm building openblas and atlas packages from source.  No splitting out the -dev stuff into a separate package, no distro-specific additions.

openblas creates [openblas.pc](https://github.com/xianyi/OpenBLAS/blob/develop/openblas.pc.in) and Debian is [adding a few more](https://sources.debian.org/src/openblas/0.3.5+ds-1/debian/).
I can't find any pkgconfig files in the atlas 3.10.3 source, so Debian is [adding their own](https://sources.debian.org/src/atlas/3.10.3-7/debian/).

> [9c77ab2](https://github.com/opencollab/arpack-ng/commit/9c77ab2252f1006544387929c0b2083129157cd6) may be a way to deal with that ?...

Won't help.  The check will always fail if the pkgconfig file doesn't exist, or (with this change) exists but the version is too old.  This is a way to specify that earlier versions of openblas/atlas are not compatible with arpack.  Probably not what you were intending since I imagine they are compatible.

> personal opinion: always prefer last updated versions + think "ahead" and use new arch/package (not target old-cray-or-archaic arch/package - the day you're stuck here: use ax_blas the "old" way specifying libs by hand...)

FindBLAS and ax_blas.m4 are basically doing the same job, even the FindBLAS documentation says the search is based on acx_blas.m4.  The checks they are doing still have value over pkg-config, they make sure that the library is usable and contains the right symbols.

fghoussen:
@turboencabulator : I agree with you !... So this PR may be closed-but-not-merged ? Too bad...

fghoussen:
By the way, one may want to keep the precision fix from this PR anyway ?... I guess would be good.

sylvestre:
So, what do we want to do it here? Closing?

fghoussen:
> So, what do we want to do it here? Closing?

Yes closing... As pc files are provided by debian only. Unfortunately, no solution here...

You may want to keep the "precision fix" from this PR (with atlas, but not with openblas or blas/lapack, CI was failing because of precision problem related to ***s**[ae]upd) : not critical, but, if you want that, let me know, I'll open a new PR for that only.

I'am closing this as there is no solution for now (no pc files => no --enable options)

Issue #195, opened 2019-02-06
icb_arpack_cpp test fails with ILP64 builds on Fedora


rathann:
## Expected behavior

All tests in the testsuite should PASS.

## Actual behavior

`icb_arpack_cpp` test FAILs with abort/core dump.

## Where/how to reproduce the problem

- arpack-ng: 3.7.0
- OS: Fedora rawhide, 64-bit arches (x86_64, aarch64, ppc64le, s390x)
- compiler: gcc-9.0.1
- environment: 
```
+ CFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'
+ export CFLAGS
+ CXXFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'
+ export CXXFLAGS
+ FFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -I/usr/lib64/gfortran/modules'
+ export FFLAGS
[...]

PR #197, opened 2019-02-07
[BUG FIX] arpack may succeed to compute more EV than expected.

fghoussen:
## Pull request purpose

fixing issue #195 ?

- [BUG FIX] arpack may succeed to compute more EV than expected.


fghoussen:
no way 1f1a1d0 could fix coverage !....

PR #196, opened 2019-02-06
[BUG FIX] ILP64 support: using debug_c and stat_c.

fghoussen:
## Pull request purpose

fixing issue #195 ?

## Detailed changes proposed in this pull request

BUG FIX: add a_int in stat_c.h and debug_c.h

PR #198, opened 2019-02-07
[BUG FIX] fix check precision which may fail with some ATLAS versions.

fghoussen:
## Pull request purpose

fixing issue #195 ?

[BUG FIX] fix check precision which may fail with some ATLAS versions.

### DIFF
--- a/TESTS/icb_arpack_cpp.cpp
+++ b/TESTS/icb_arpack_cpp.cpp
@@ -27,7 +27,7 @@ void diagonal_matrix_vector_product(Real const* const x, Real* const y) {
 }
 
 template<typename Real>
-void real_symmetric_runner() {
+void real_symmetric_runner(double const & tol_check) {
   a_int const N = 1000;
   a_int const nev = 9;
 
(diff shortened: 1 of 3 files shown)


TESTS/icb_arpack_cpp.cpp
```diff
@@ -177,8 +177,8 @@ int main() {
   sstats_c();
 
   // arpack without debug
-  real_symmetric_runner<float>();
-  real_symmetric_runner<double>();
+  real_symmetric_runner<float>(1);
+  real_symmetric_runner<double>(1.e-05);
 
   int nopx_c, nbx_c, nrorth_c, nitref_c, nrstrt_c;
```
rathann: This fails to apply on top of #196, which changes `int` to `a_int` here.

fghoussen: #196, #197 and #198 must be tested separately : for each of them, you need to apply all commits of each PR (= 1 branch at my side). Tests must be done separately as they modify the same files: they indeed do conflict, I'll have to rebase them one after the others before they could be merged.

I did 3 PRs as each of them test a different idea/reason. 

Maybe more simple for you (?) to clone my repo (https://github.com/fghoussen/arpack-ng) and, then, switch between branches (issue95 is #196, check is #197 and floatPrecision is #198)

fghoussen: I rebased + pushed-force #196 : each PR is one commit now => may be easier to test for you

fghoussen:
BTW, this was in connection with #186 (was not merged).

Issue #210, opened 2019-05-13
ISO_C_Binding CMake configuration fails


zbeekman:
## Expected behavior

Build successfully when `-DICB=ON` with CMake

## Actual behavior

```
-- Looking for Fortran sgemm - found
-- Found BLAS: /usr/local/opt/openblas/lib/libopenblas.dylib
-- Found MPI_C: /usr/local/lib/libmpi.dylib (found version "3.1")
-- Found MPI_CXX: /usr/local/lib/libmpicxx.dylib (found version "3.1")
-- Found MPI_Fortran: /usr/local/lib/libmpifort.dylib (found version "3.1")
-- Found MPI: TRUE (found version "3.1")
CMake Error at CMakeLists.txt:184 (check_symbol_exists):
  Unknown CMake command "check_symbol_exists".
```

## Where/how to reproduce the problem

- arpack-ng: Master and 3.7.0
- OS: mac, but shouldn't make a difference
- compiler: gcc 9.1 but shouldn't make a difference
- environment: Default Homebrew environment (you can see it with `brew --env`)
- configure: `cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/arpack/3.7.0_2 -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -Wno-dev -DBLA_VENDOR:STRING=OpenBLAS -DMPI:BOOL=ON -DEXAMPLES:BOOL=ON -DICB:BOOL=ON -DICBEXMM:BOOL=ON`
- input data N/A
- CMake version: 3.14.3

## Steps to reproduce the problem
- Try configuring with CMake with the flags above. In fact, just setting `-DICB=ON` should trigger the error

## Error message

```
CMake Error at CMakeLists.txt:184 (check_symbol_exists):
  Unknown CMake command "check_symbol_exists".
```

## Traces

N/A

## Callstack

N/A

## Notes, remarks

N/A, see details above

PR #211, opened 2019-05-13
Fix CMake bug when enabling ISO_C_binding support

zbeekman:
## Pull request purpose

fixing issue #210

You can't access `check_symbol_exists` without including the appropriate intrinsic module first. If you don't make this change then the ISO_C_Binding build breaks with modern CMake.


sylvestre:
thanks
please also update the changelog !

PR #220, opened 2019-07-07
TravisCI: fix ILP64 job. No need for security.debian.org.

fghoussen:
Try fix for broken ILP64 update.


fghoussen:
OK. So this one must be merged first

Issue #214, opened 2019-05-26
Automatic linking with openblas


cpviolator:
I would like to build arpack-ng as a dependency to a higher level library. Arpack-ng will automatically link to an installed openblas shared library on some systems. This openblas library is not linked in the higher level library, causing link errors at build time with Make.

How can I pass a CMake flag to arpack-ng to block this linking to openblas?

fghoussen:
If hll (higher level lib) does not need to link to openblas, but, does need to link to arpack-ng (which links to openblas), then hll should not see openblas : all should be fine. Are you sure hll does not link to openblas : what does ldd says ? Compile / link traces ? Error message ?

> How can I pass a CMake flag to arpack-ng to block this linking to openblas so that ARPACK builds it own blas libs?

Not sure arpack-ng builds it's own blas. I would say arpack-ng uses any blas it finds... And fails compiling if no blas found.

cpviolator:
At the moment, we download and build ARAPCK from the Rice university source, these are the relevant lines in the makefile:
```
FC      = /usr/local/mvapich2-2.2-gcc-6.3-cuda-9.2/bin/mpifort
FFLAGS  = -I/usr/local/mvapich2-2.2-gcc-6.3-cuda-9.2/include -fPIC

# Set the parallel fortran compiler
MPILIBS = -L/usr/local/mvapich2-2.2-gcc-6.3-cuda-9.2/lib/libmpifort.so
PFC     = /usr/local/mvapich2-2.2-gcc-6.3-cuda-9.2/bin/mpifort
PFFLAGS = -I/usr/local/mvapich2-2.2-gcc-6.3-cuda-9.2/include -fPIC
```
There are lines in the makefile to link to BLAS and LAPACK, which we do not use. If we link ARPACK to the installed OpenBLAS shared library, we must also link to the openBLAS shared library in the hll. We wish to avoid this.

fghoussen:
Does this fix the problem ? (rebuild from scratch)

```
>> git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 15f9bb8..884d17b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,8 +257,8 @@ set(arpackutil_STAT_SRCS
 
 
 if (MPI)
-    file(GLOB parpackutil_STAT_SRCS ${arpack_SOURCE_DIR}/PARPACK/UTIL/MPI/*.f)
[...]

PR #221, opened 2019-07-13
[BUG FIX] Fix pc/cmake files.

fghoussen:
## Pull request purpose

fixing bug: autotools get eigen with -I flavour, flake don't.


configure.ac
```diff
@@ -242,7 +242,7 @@ AC_CONFIG_FILES([arpack$LIBSUFFIX.pc:arpack.pc.in], [], [LIBSUFFIX="$LIBSUFFIX"]
 AC_SUBST([CMAKE_INSTALL_PREFIX], ["$prefix"])
 AC_SUBST([BLAS_LIBRARIES], ["$BLAS_LIBS"])
 AC_SUBST([LAPACK_LIBRARIES], ["$LAPACK_LIBS"])
-AC_SUBST([EIGEN3_INCLUDE_DIR], ["$EIGEN3_CFLAGS"])
+AC_SUBST([EIGEN3_INCLUDE_DIR], ["${EIGEN3_CFLAGS/-I/}"])
```
fghoussen: @turboencabulator : do you think there is a more portable / autotools way to do this ?

turboencabulator: @fghoussen That's a bashism, won't work on all platforms (particularly BSD, they've already had plenty of problems with bash).  I think this and #216 need to be reverted, they're going to cause more problems than they solve.  If I understand #214 right, underlinking is either the goal or the problem, and this will force overlinking.

fghoussen: @turboencabulator : thanks for the feedback.
- if you think some patchs/bashism need to be reverted, could you PR some replacement ?
  At my side on debian/testing + bash #216 was indeed needed to compile/link. 
- Would really be good to add Free BSD jobs in CI...
  Too bad that docker can't make it (different kernels as far as I understand)

turboencabulator: @fghoussen Yeah I'll send a PR.  Can you elaborate on your testing?  Were you trying to use a static or a shared lib?  Ideally the output from pkg-config looks something like this, and if you're using a static lib then it'll give you the extra flags you need to link (making this whole change unnecessary).
```
$ pkg-config --libs arpack
-larpack
$ pkg-config --libs --static arpack
-larpack -llapack -lblas
```

fghoussen: @turboencabulator : thanks. Tried both linking.

Issue #224, opened 2019-07-24
Data type issues in ICB interfaces


pv:
## Expected behavior

The ICB wrappers should correctly pass arguments such as `rvec` to
the Fortran routines.

## Actual behavior

`logical(kind=c_bool)` and `logical` are not compatible types on e.g. gcc/gfortran 9,
in particular having different sizes (1 byte vs. 4 bytes).

As a consequence, it breaks passing `rvec=false` to the Fortran routines, because
the larger data type accesses some garbage on stack and becomes nonzero.

## Where/how to reproduce the problem

- arpack-ng: 8a46621 
- OS: fedora
- compiler: gcc/gfortran 9.1.1
- configure: ./configure --enable-icb

## Steps to reproduce the problem

You can step from `dseupd_c` to `dseupd` in `icb_arpack_c` using gdb and see `rvec` is not correctly passed.

Or, like so:

```shell
set -e

git clone --depth 1 https://github.com/opencollab/arpack-ng.git tmp
cd tmp

# add print+exit statement to beginning of dseupd.f, just for simple illustration
sed -i -e "s/msglvl = mseupd/write(*,*) 'dseupd rvec =', rvec\n      stop/" SRC/dseupd.f

# call via ICB
cat <<__EOF__ > testbug.c
#include <stdbool.h>
#include "arpack.h"
int main() {
    bool rvec = false;
    dseupd_c(rvec, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
    return 0;
}
__EOF__

./bootstrap
./configure --enable-icb
make -j6

gcc -IICB -I. -o testbug testbug.c SRC/.libs/libarpack.a -llapack -lblas -lgfortran -lm

echo "The following should print rvec = .false."
./testbug
```
which prints here
```
The following should print rvec = .false.
 dseupd rvec = T
```

fghoussen:
OK, far from home. I'll try to have a look when possible. Did you try with several compiler / version ?

fghoussen:
Pb fixed: closing issue.

PR #222, opened 2019-07-15
[API] arpackSolver: API change.

fghoussen:
## Pull request purpose

arpackSolver: API change.

PR #225, opened 2019-08-03
[BUG FIX] ICB using rvec/select: rvec/select turned to integer

fghoussen:
bool should be, but, is not always supported (depend on compiler, options).

## Pull request purpose

fixing issue #224 

## Notes

#224 is true bug, even if the code as-is _should_ work : it does not as compilers may not support it / or need specific options (which would complicate the configure/cmake...).
See:
- https://software.intel.com/en-us/forums/intel-fortran-compiler-for-linux-and-mac-os-x/topic/594856
- https://gist.github.com/certik/9744747

Simple solution: replace bool with integer (= use basics types only)


fghoussen:
no clue with CI breaks !? mirror pb ? pushing another (missing) commit anyway.

sylvestre:
I am worried you are changing the profile of pretty old functions.
I would prefer another solution

fghoussen:
Never changed any old function !

Changes only apply to ICB function (= "new" functions which exist only to wrap old ones) and C/C++ headers. Even in binary: old symbols still exist as they always have (*upd_), but, new ICB symbols are added (*upd_c).

Seems CI breaks because of mirror problems

fghoussen:
fixed-up and pushed force the exact same modifs: see if mirrors are OK now.

fghoussen:
BTW, as-is (i.e. without this PR) types are wrong (OK initially but changes have been made along the way !): checkout `select`, it is an `a_int` in arpack.h/hpp but `logical` (not integer) in icb wrappers (type mismatch : this is actually a bug nest and should not compile - I guess it compiles as there is a possible cast between int and bool).

This is necessary and does not change/touch at all any "old" functions. All what is related to ICB is "just" typing : associating types to variables, this is done in wrappers that call old-original-untouched functions afterwards. ICB is just a gateway.

fghoussen:
If you want to avoid this, you still need to:
- make `select`and `rvec` the same type (int or bool) in both arpack.h/hpp and icb wrappers : this means you need a PR which will look like this one anyway (changes in .h/hpp and icb wrappers - old functions untouched)
- if you want to keep `logical` instead of `integer`, you could but you need to look for exotic compile options (allow logical if icb - checkout url I gave above) for all possible compilers in both autotools/cmake... With is tricky, difficult to test, and likely bug-prone...

This is the simplest way: integer are always supported as they are native types. ICB wrappers make the convertion and, then, the call the "old" functions with correct types

fghoussen:
> I am worried you are changing the profile of pretty old functions.
@sylvestre : what do you mean by profile ?

sylvestre:
This
image

fghoussen:
This is  the icb wrapper: this __is__ the bug.

sylvestre:
ok, thanks
will this require a rebuild of libraries using the icb wrapper?

fghoussen:
> will this require a rebuild of libraries using the icb wrapper?

Yes (no choice : type mismatch in F90 / C headers) and may break the build.
But ICB is not such a well-known option: likely not used that much !... Better to fix that sooner than later.

pv:
You don't need to change the data types in the ICB interface. They can be retained as `logical(kind=c_bool)`. This needs the parameter conversion code, but it's essentially the same as what was already done in this PR for `aint`.

pv:
The ICB interface files also should contain the `begin interface ... end interface` declarations for the ARPACK routines called. This way you will get a compiler error instead of runtime failures if the data type declarations do not match.

fghoussen:
@pv: I let you the hands on it !

> You don't need to change the data types in the ICB interface. 

Yes you'll need it as there is already-as-is a type mismatch : in arpack.h `rvec` (resp. `select`) is `bool` (resp. `a_int`), but, in icba*.F90 `rvec` (resp. `select`) is `logical` (resp. `logical`). In any cases, these must be consistent : `bool` must match `logical`, `a_int` must match `integer` (as-is, OK for `rvec`, but, KO for `select`). Note: the initial problem was about `rvec` (only) which is OK (no type mismatch), and then, I realized there was also a (mute) problem with `select`.

If you keep `logical` : seems (not tested - no ifort available here) that if you use ifort you need to add specific compiler options `-fpscomp logical` (https://gist.github.com/certik/9744747). If so, then you need to impact autotools and cmake to find these specific options depending on each given specific compiler (gnu, intel, pgi, ...) : this may not be an easy game ! Getting back to native types looked easier to me.

> The ICB interface files also should contain the begin interface ... end interface declarations for the ARPACK routines called. This way you will get a compiler error instead of runtime failures if the data type declarations do not match.

I guess you are right ! At the very beginning of this, I found no way to do this using `interface`. So I started the most simple as possible (indeed, no check). Adding `interface` would be a good thing (get a compiler error if data type mismatch) : maybe easier to do that now from here than from scratch. If you succeed in this : PR will be appreciated.

fghoussen:
Untested but check out the answer of 'Repeat Offender' here : https://software.intel.com/en-us/forums/intel-fortran-compiler-for-linux-and-mac-os-x/topic/594856

> You need the switch -fpscomp logicals to get the behavior that semi-matches gcc. There is a bug in most Fortran compilers in that they return C_BOOL /= -1, and ifort is one of such buggy compilers. The matching type, _Bool, is an oddball underspecified type even in C99, and it pretty much breaks LOGICAL variables in Fortran when the compiler tries to match it. My recommendation is to forget about trying to interoperate with _Bool and instead use int on the C side just like all actual C programs do. On the Fortran side the appropriate type is INTEGER(C_INT) and you can check whether it is zero or not just as C programs do.

fghoussen:
For me, this PR is good to go (if agreed with others).
In a second time, if @pv could add `interface` would be great.

pv:
If you want to break backward compatibility to work around potential compiler bugs (which may be fixed today --- the links cited are a few years back), that's your call. I don't use the icb interfaces myself, but I pointed out a simple way to avoid problems like this in future.

fghoussen:
```
>> git diff
diff --git a/SRC/icbacn.F90 b/SRC/icbacn.F90
index cf53853..7425d16 100644
--- a/SRC/icbacn.F90
+++ b/SRC/icbacn.F90
@@ -1,5 +1,7 @@
 ! icba : iso_c_binding for arpack
 
+interface
+
 subroutine cnaupd_c(ido, bmat, n, which, nev, tol, resid, ncv, v, ldv,&
                     iparam, ipntr, workd, workl, lworkl, rwork, info) &
                     bind(c, name="cnaupd_c")
@@ -79,3 +81,5 @@ subroutine cneupd_c(rvec, howmny, select, d, z, ldz, sigma, workev,  &
               bmat, n, which, nev, tol, resid, ncv, v, ldv,  &
               iparam, ipntr, workd, workl, lworkl, rwork, info)
 end subroutine cneupd_c
+
+end interface

>> make
icbacn.F90:29:63:

               iparam, ipntr, workd, workl, lworkl, rwork, info)
                                                               1
Error: Unexpected CALL statement at (1) in INTERFACE body

̀>> git diff
diff --git a/SRC/icbacn.F90 b/SRC/icbacn.F90
index cf53853..05a7652 100644
--- a/SRC/icbacn.F90
+++ b/SRC/icbacn.F90
@@ -1,5 +1,9 @@
 ! icba : iso_c_binding for arpack
 
+module
+
+interface
+
 subroutine cnaupd_c(ido, bmat, n, which, nev, tol, resid, ncv, v, ldv,&
                     iparam, ipntr, workd, workl, lworkl, rwork, info) &
                     bind(c, name="cnaupd_c")
@@ -79,3 +83,7 @@ subroutine cneupd_c(rvec, howmny, select, d, z, ldz, sigma, workev,  &
               bmat, n, which, nev, tol, resid, ncv, v, ldv,  &
               iparam, ipntr, workd, workl, lworkl, rwork, info)
 end subroutine cneupd_c
+
+end interface
+
+end module

>> make
icbacn.F90:3:6:

 module
      1
Error: Invalid character in name at (1)
icbacn.F90:31:63:

               iparam, ipntr, workd, workl, lworkl, rwork, info)
                                                               1
Error: Unexpected CALL statement at (1) in INTERFACE body

>> git diff
diff --git a/SRC/icbacn.F90 b/SRC/icbacn.F90
index cf53853..777cc78 100644
--- a/SRC/icbacn.F90
+++ b/SRC/icbacn.F90
@@ -1,5 +1,13 @@
 ! icba : iso_c_binding for arpack
 
+module
+
+interface
+  module procedure cnaupd_c, cneupd_c
+end interface
+
+contains
+
 subroutine cnaupd_c(ido, bmat, n, which, nev, tol, resid, ncv, v, ldv,&
                     iparam, ipntr, workd, workl, lworkl, rwork, info) &
                     bind(c, name="cnaupd_c")
@@ -79,3 +87,5 @@ subroutine cneupd_c(rvec, howmny, select, d, z, ldz, sigma, workev,  &
               bmat, n, which, nev, tol, resid, ncv, v, ldv,  &
               iparam, ipntr, workd, workl, lworkl, rwork, info)
 end subroutine cneupd_c
+
+end module

>> make
icbacn.F90:3:6:

 module
      1
Error: Invalid character in name at (1)
icbacn.F90:6:18:

   module procedure cnaupd_c, cneupd_c
                  1
Error: MODULE PROCEDURE at (1) must be in a generic module interface
icbacn.F90:13:44:

                     bind(c, name="cnaupd_c")
                                            1
Error: No binding name is allowed in BIND(C) at (1)

>> gcc --version
gcc (Debian 8.3.0-6) 8.3.0

```

fghoussen:
@sylvestre : BTW, if you merge that PR, #222 will be broken.
You should merge this PR first (or replace it by another you would prefer).

Issue #233, opened 2020-01-17
Another error when building with LLVM on MacOS


cesss:
I tried to build the 3.7.0 release on MacOS with LLVM and got errors related to headers for complex numbers. Then, after searching, I found this commit, more recent than the 3.7.0 release, so I tried to build the master branch instead.

But it failed with another error later, and also related to complex numbers headers:

```
In file included from /Users/src/arpack-ng/2020-01-17/PARPACK/TESTS/MPI/icb_parpack_cpp.cpp:14:
/Users/src/arpack-ng/2020-01-17/ICB/parpack.hpp:138:42: error: use of undeclared identifier '_Complex_I'
                      std::real(sigma) + _Complex_I * std::imag(sigma),
                                         ^
/Users/src/arpack-ng/2020-01-17/ICB/parpack.hpp:176:42: error: use of undeclared identifier '_Complex_I'
                      std::real(sigma) + _Complex_I * std::imag(sigma),
                                         ^
2 errors generated.

```

cesss:
Actually, researching it a bit more, it seems it's not a LLVM/MacOS specific issue, but that `parpack.hpp` is a C++ header, so it shouldn't use` _Complex_I`, which is a C complex number (read the StackOverflow accepted answer to [this question](https://stackoverflow.com/questions/44203873/how-to-compile-c-complex-numbers-with-c11))

cesss:
I tried doing in `parpack.hpp` the same modifications as were done to `arpack.hpp` and the build is successful, but however, when doing `make test`, the test 10/11 fails, and the test 11/11 seems to be stuck in an endless loop (two processes at 100% CPU, that I had to manually kill).

Curiously, the two tests that fail are the C++ tests:

```
make test
Running tests...
/Users/cesss/CMake/3.15.5/bin/ctest --force-new-ctest-process 
Test project /Users/cesss/build/arpack-ng/2020-01-17
      Start  1: dnsimp_tst
 1/11 Test  #1: dnsimp_tst .......................   Passed    0.02 sec
      Start  2: bug_1323_tst
 2/11 Test  #2: bug_1323_tst .....................   Passed    0.00 sec
      Start  3: bug_58_double_tst
 3/11 Test  #3: bug_58_double_tst ................   Passed    0.00 sec
      Start  4: bug_79_double_complex_tst
 4/11 Test  #4: bug_79_double_complex_tst ........   Passed    0.00 sec
      Start  5: bug_142_tst
[...]

PR #234, opened 2020-01-18
[TravisCI] Add osx job

fghoussen:
## Pull request purpose

fixing issue #233 ?...
@cesss : you tell me what to do next !... not mac-based

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ services:
 
 stages:
   # order stages
+  - name: osx
   - name: precise
   - name: trusty
   - name: xenial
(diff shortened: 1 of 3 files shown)


fghoussen:
same problem than with windows build !... tried to brew some package around ?... let's see

fghoussen:
@cesss : what do you say ?

cesss:
> @cesss : what do you say ?

Unfortunately I have no experience at automating builds and checks with travis --that's still in my TODO list 😄

On my Mac, it's what I said at the issue: this change in parpack.hpp does make the ICB build complete successfully, but then the two C++ tests fail (well, one of them fails, and the other gets stuck in an endless loop)

fghoussen:
> I have no experience at automating builds and checks with travis

@cesss : Look at commits: now you have it. Not much about that anyway, it's about that : how to you brew you mac to get a clean env where to build arpack ? FC compiler ? packages ? brew options / conflicts ? ... I never had a mac so this is difficult for me to know

sylvestre:
Maybe look at similar projects, how they are doing it!

fghoussen:
> Maybe look at similar projects, how they are doing it!

Not clear / easy : step-by-step blind pushing...

fghoussen:
@cesss : AFK, do you reproduce crashes when using `-D EXAMPLE=ON` ? (`[cz]ndrv*` are full fortran AFAIK, _not_ C++ related) I suspect compilers getting in each other's way : I ask for `llvm` but get `gcc`, how to make `[cf]lang` as default compilers on osx ?

fghoussen:
Splitting this in 2 PRs : code fix + CI job

fghoussen:
@cesss : do you reproduce the crash for the full-fortran `cnbdr3_ex` test ? can you get a stack ? 
seems impossible to get a stack from travis-ci box : sounds like the problem could be in lapack...

fghoussen:
OK, does not work: stopping here. If somebody has an idea or can get a stack on osx, speak up !...

fghoussen:
@sylvestre : 

1. d5ff2d7f99 fix build on osx when using autotool

2. if you accept (?) full build but only partial testing, you may get a green light on osx... not "great" but "better than nothing" (personal opinion) : you can at least test build is OK (which is not without d5ff2d7f99 for autotool) and at least a few tests

@cesss : can you get a stack on osx for failed tests like the full-fortran `cnbdr3_ex` for example ?

.travis.yml
```diff
@@ -39,6 +40,38 @@ stages:
...
+    osx_image: xcode10.1
+    before_install:
+    - brew list -1 | while read line; do brew unlink $line; done;
+    - brew list gcc      || brew install gcc
+    - brew list cmake    || brew install cmake
+    - brew list openblas || brew install openblas
+    - brew list lapack   || brew install lapack
+    - brew list openmpi  || brew install openmpi
+    - brew list -1 | while read line; do brew link --overwrite $line; done;
+    # full build but only partial testing as some tests crash on osx (no stack: if you get one, open an issue https://github.com/opencollab/arpack-ng/issues)
```
fghoussen: @sylvestre : caution, only partial testing

.travis.yml
```diff
@@ -39,6 +40,38 @@ stages:
...
+    before_install:
+    - brew list -1 | while read line; do brew unlink $line; done;
+    - brew list gcc      || brew install gcc
+    - brew list autoconf || brew install autoconf
+    - brew list automake || brew install automake
+    - brew list openblas || brew install openblas
+    - brew list lapack   || brew install lapack
+    - brew list openmpi  || brew install openmpi
+    - brew list -1 | while read line; do brew link --overwrite $line; done;
+    # full build but only partial testing as some tests crash on osx (no stack: if you get one, open an issue https://github.com/opencollab/arpack-ng/issues)
```
fghoussen: @sylvestre : caution, only partial testing here too

PR #239, opened 2020-02-19
?vout expects a vector, so make sure to pass a vector

dschwoerer:
## Pull request purpose

Make arpack compatible with gcc 10

replacing scalar variables with vectors by enclosing `[]` around it, as described [in the porting guide](https://gcc.gnu.org/gcc-10/porting_to.html)

With these fixes, arpack compiles and the tests run as well with gcc 10.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,9 @@ arpack-ng - 3.8.0
 [ Kyle Guinn ]
  * [BUG FIX]: fix 'eval: Syntax error: "(" unexpected' error at build time.
 
+[ David Schwörer ]
+ * [BUG FIX] port to gcc 10
+
  -- Sylvestre Ledru <sylvestre@debian.org> Mon, 1 April 2019 16:24:00 +0100
 
 arpack-ng - 3.7.0
--- a/CHANGES
+++ b/CHANGES
@@ -21,7 +21,7 @@ arpack-ng - 3.8.0
  * [BUG FIX]: fix 'eval: Syntax error: "(" unexpected' error at build time.
 
 [ David Schwörer ]
- * [BUG FIX] port to gcc 10
+ * Support of gfortran 10
 
  -- Sylvestre Ledru <sylvestre@debian.org> Mon, 1 April 2019 16:24:00 +0100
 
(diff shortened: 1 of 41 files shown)


sylvestre:
@fghoussen wdyt? thanks

sylvestre:
@dschwoerer could you please also update the changelog? thanks!

CHANGES
```diff
@@ -20,6 +20,9 @@ arpack-ng - 3.8.0
 [ Kyle Guinn ]
  * [BUG FIX]: fix 'eval: Syntax error: "(" unexpected' error at build time.
 
+[ David Schwörer ]
+ * [BUG FIX] port to gcc 10
```
sylvestre:
```suggestion
 * Support of gfortran 10
```

fghoussen:
@sylvestre : not aware of that.

@dschwoerer : could understand it's a necessary change (which must be done) but to me, first, it's necessary to check if :

-  it does not break compilation with other compilers (icc/ifort/clang...) : did you check that ?

- you add 1 or 2 other build in CI dedicated to test : 

   1.  build with gcc-10 (apt-get install gcc-10)

   2. build with clang (apt-get clang flang)

dschwoerer:
I tried `icc/ifortran` and `make all check` failed, but `make ; make all; make check` worked.

`flang` seems to be only for llvm 7, but fedora ships llvm 10. ubuntu:latest doesn't seem to have flang either (I tried `apt-get update; apt-get install flang`)

Will add an test with gcc-10, not sure how to do llvm ...

I noticed I only ported the serial version. The parallel isn't that hard to get compiling, but then some tests fail. As far as I understand `[]` creates a temporary array, which is fine for printing, but not for receiving (MPI) data. So that will need some more effort ...

fghoussen:
@dschwoerer  : for ` gcc-10 `  you likely could use debian folowwing example https://github.com/opencollab/arpack-ng/blob/3dd2d5fe97ed40759b3b175d4a1c27ebcd9b0921/.travis.yml#L231 where I see `  gcc-10 `  packages (not tested). Fro `  flang/clang `  you may use debian another OS (ubuntu or whatever you are familiar with). Unfortunately, for `  icc/ifort/pgi `  not way to test with open-sourced CI...

Issue #230, opened 2019-11-30
icb_arpack_cpp test FAILs on ppc64le on Fedora


rathann:
## Expected behavior

All tests in the testsuite should PASS.

## Actual behavior

icb_arpack_cpp test FAILs on ppc64le on Fedora.

## Where/how to reproduce the problem

- arpack-ng: 3.7.0
- OS: Fedora 32/rawhide
- compiler: gcc 9.2.1
- environment:
```
+ CFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mcpu=power8 -mtune=power8 -fasynchronous-unwind-tables -fstack-clash-protection'
+ export CFLAGS
+ CXXFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mcpu=power8 -mtune=power8 -fasynchronous-unwind-tables -fstack-clash-protection'
+ export CXXFLAGS
+ FFLAGS='-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mcpu=power8 -mtune=power8 -fasynchronous-unwind-tables -fstack-clash-protection -I/usr/lib64/gfortran/modules'
+ export FFLAGS
[...]

PR #244, opened 2020-02-20
[TravisCI] Add ILP64 build with cmake.

fghoussen:
## Pull request purpose

Add cmake ILP64 build. "Should" fail because of ICB : this is #230... If KO, I'll suppress ICB temporarily to keep testing the most possible things step-by-step...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -292,7 +292,11 @@ jobs:
        git status                                                                                                             && \
        git log -2                                                                                                             && \
        mkdir build && cd build                                                                                                && \
-       BLA_VENDOR='Intel10_64ilp' BLA_PREFER_PKGCONFIG=1 cmake -D INTERFACE64=ON -D EXAMPLES=ON -D ICB=ON ..                  && \
+       FFLAGS='-I/usr/include/mkl'                                                                                               \
+       FCFLAGS='-I/usr/include/mkl'                                                                                              \
+       BLAS_LIBRARIES='-Wl,--no-as-needed -lmkl_sequential -lmkl_core -lpthread -lm -ldl'                                        \
+       LAPACK_LIBRARIES='-Wl,--no-as-needed -lmkl_sequential -lmkl_core -lpthread -lm -ldl'                                      \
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -225,6 +225,46 @@ jobs:
       sudo docker cp -a ${TRAVIS_BUILD_DIR} mobydick:/tmp                                                               \
       &&                                                                                                                \
       sudo docker start -a mobydick
+  - stage: interface64
```
sylvestre: same as the other PR, lot of code duplication here :/

fghoussen: No sure to get what you mean: AFAIK, in yml, no way to "factorize" code or "make functions", or, I just don't know how to do that. You need to add job stage by stage as yml grammar impose it, or, you could put all jobs in one stage but then all config are mixed (difficult to debug when pb shows up) and huge-logs blow up (can't access traces when pb show up)

fghoussen:
Yet another problem ?!... CI freezes for ilp64 cmake : no idea why ? Trying static build

fghoussen:
Great : BLA_STATIC finds .so ?!... Build OK but run freezes ?!...

fghoussen:
Will try a few tricks and give up if this is too tricky to get done : could be endless...

fghoussen:
WTF ? 3dd5ac6 cancels out 72cc5e3 CI ?!... can restart 72cc5e3 jobs. will test one after the other

.travis.yml
sylvestre: most of the operations should be moved into a script
link "setup-env.sh" and called with a bash -v

fghoussen:
Cmake still "decide" to link with ` *lp64.so ` not **i**lp64 ?!... Try 2 more tricks and give up....

fghoussen:
` BLA_VENDOR='Intel10_64ilp' cmake -D INTERFACE64=ON -D EXAMPLES=ON -D ICB=ON -D MPI=ON .. ` just  work fine on my debian box but fails on CI (build OK, running tests freeze ) ?! Will try a few more trick and give up...

fghoussen:
WTF ?! Endless burden : giving up.

PR #246, opened 2020-02-25
[BUG FIX - architecture] fix arpackdef.h: must not be included by F77/F90.

fghoussen:
This should fail.

PR #247, opened 2020-02-26
[BUG FIX - ICB] iparam/ipntr sizes may change depending on cases.

fghoussen:
And this is a silent bug...

### DIFF
--- a/EXAMPLES/MATRIX_MARKET/arpackSolver.hpp
+++ b/EXAMPLES/MATRIX_MARKET/arpackSolver.hpp
@@ -10,6 +10,8 @@
 #include <chrono>
 #include <iomanip> // setw.
 #include <cassert>
+#include <vector>
+#include <type_traits> // is_same.
 
 #include <Eigen/Sparse>
 #include <Eigen/IterativeLinearSolvers>
@@ -670,6 +672,21 @@ class arpackSolver {
       }
     };
 
+    int initPointerSize(a_int & iparamSz, a_int & ipntrSz, string const & aeupd) {
+      iparamSz = 11; ipntrSz = 14;
+      if (aeupd == "aupd") {
+        if (is_same<RC, double>::value && symPb) ipntrSz = 11;
+        if (is_same<RC,  float>::value && symPb) ipntrSz = 11;
+        return 0;
+      }
+      else if (aeupd == "eupd") {
+        if (is_same<RC, double>::value && symPb) {iparamSz = 7; ipntrSz = 11;}
+        if (is_same<RC,  float>::value && symPb) {iparamSz = 7; ipntrSz = 11;}
+        return 0;
+      }
+      return 1;
+    };
+
     int solve(EM const & A, EM const * B, SLV & solver) {
       if (!stdPb && !B) {cerr << "Error: generalized problem without B" << endl; return 1;}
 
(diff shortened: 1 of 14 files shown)


sylvestre:
Could you please add a test which reproduce the bug and make sure it doesn't happen anymore?
Also please update the changelog
thanks

fghoussen:
> Could you please add a test which reproduce the bug and make sure it doesn't happen anymore?

Not really. Always / often happening :  allocating more memory than needed (silent-not-hurting bug). The real question is : if 7d341a9 is OK (on LP64), then, could this fix #230 which seems to show up only on **I**LP64 ? (fix is related to ` integer` )

fghoussen:
All OK... So one can try to make the problem "reappear" here after rebase on #246 to benefit from c170c16 and check out if this could fix what looks like #230 ?

fghoussen:
OK, here : this is a silent-no-hurting bug. This should _not_ create any problem : it's just that you may sometimes  allocate a bit more memory than you need (14/11 integers instead of 11/7). You always allocate more : so there should be no worry here. 

At first, I suspected this may be the cause of #230 (seems to be **I**LP64 specific) as it's related to ` iparam / ipntr ` which are arrays of ` integer `. Seems my suspicion was wrong...


I let this PR aside for now. No problem if not merged.

fghoussen:
After looking back at this, this should preferably be merged :  

- I realized that in many cases ` ipntr ` size is actually **increased** in ` icb*.F90 ` ! (need **more** memory than expected) 

- But in ` icb*.F90 ` , the sizes are likely muted variables / not-used information (not sure) and are probably "just not checked anyway" (what is checked by compiler is the _type_, not sure if the size is checked too)

- In pratical cases, using the same size for ` iparam / ipntr ` is more convenient (even when you allocate more than needed) : hunting all cases in the repo to "fix" this would be endless and error-prone... So I don't want to go into this. But maybe you could keep the "fix" in ` arpackSolver.hpp ` (as this one is easy : this is the **one** place where you always pass by)

fghoussen:
Rebasing on master as we know now this is no big bug and can't be related to **I**LP64 problems.

EXAMPLES/MATRIX_MARKET/arpackSolver.hpp
```diff
@@ -670,6 +672,21 @@ class arpackSolver {
...
     };
 
+    int initPointerSize(a_int & iparamSz, a_int & ipntrSz, string const & aeupd) {
+      iparamSz = 11; ipntrSz = 14;
+      if (aeupd == "aupd") {
+        if (is_same<RC, double>::value && symPb) ipntrSz = 11;
+        if (is_same<RC,  float>::value && symPb) ipntrSz = 11;
+        return 0;
+      }
+      else if (aeupd == "eupd") {
```
sylvestre:
```suggestion
        return 0;
      }
      if (aeupd == "eupd") {
```

The else is useless as we have a return just above

fghoussen: Sure

sylvestre: fyi, you could just have press the "commit suggestion" button ;)

fghoussen: OK, now I know!

PR #248, opened 2020-02-26
pyarpack: try to use python2 and apt-get packages on CI.

fghoussen:


fghoussen:
Should fail. Don't merge !

fghoussen:
@sylvestre : for CI, the most easy thing would be to use python2 !... :D (that's why I started here)

fghoussen:
Tests fail because ` libboost-python-dev libboost-numpy-dev ` installed by ` apt-get ` are **not** built for python **3** (they are for python 2)

fghoussen:
OK, as I remembered, if you use a boost-python / boost-numpy which is **not** built for the python version you use, you get error at run-time of this kind (depending on what you need some symbols may not be defined - https://stackoverflow.com/questions/25188861/libboost-python3-so-1-56-0-undefined-symbol-pyclass-type) :
```
test 90

    Start 90: pyarpackSparseBiCGDiag_tst

90: Test command: /usr/bin/python3 "pyarpackSparseBiCGDiag.py"

90: Environment variables: 

90:  PYTHONPATH=/tmp/arpack-ng/build/lib:

90: Test timeout computed to be: 9.99988e+06

90: Traceback (most recent call last):

90:   File "pyarpackSparseBiCGDiag.py", line 5, in <module>

90:     from pyarpack import sparseBiCGDiag as pyarpackSlv

90: ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1: undefined symbol: PyClass_Type

1/1 Test #90: pyarpackSparseBiCGDiag_tst .......***Failed    0.12 sec
```

sylvestre:
> 90: ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1: undefined symbol: PyClass_Type

Seems to be the cause of the issue.

fghoussen:
If you want to use pre-built packages on CI, you need python 2 (this is what I did at first : see fba7983 that should be OK if I did no mistake - actually this is why I gave a try to ubuntu:eoan but same problem : boost-python is built for python**2**). @sylvestre : this is your decision to close or merge this PR.

fghoussen:
>>    90: ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1: undefined symbol: PyClass_Type

> Seems to be the cause of the issue.

yeap , look solution of https://stackoverflow.com/questions/25188861/libboost-python3-so-1-56-0-undefined-symbol-pyclass-type => boost-python was not built for python3

sylvestre:
Sure, I understood your point. But it is something that we can fix. We can report a bug in Debian to get that fixed.
The package HAS to support Python 3 and no longer Python 2 or it will be removed from the archive.

fghoussen:
> Sure, I understood your point. But it is something that we can fix. We can report a bug in Debian to get that fixed.

Not sure here : my understanding is that this "bug" (undefined symbols) is "normal / expected" as python**2** and python**3** are different for deep underground reasons (codebytes handling - and possibly many more stuffs - if I remember well). I'am not sure we could call that a "bug" and ask for a "fix" : this is is "just" different (to me) because "just" python2 is meant to be different from python3... 

> The package HAS to support Python 3 and no longer Python 2 or it will be removed from the archive.

Yes, you are correct. We see here that "today" the packages ` libboost-python-dev libboost-numpy-dev ` are available on bionic/eoan **and** were built for python**2**. I believe that in a "near future", these packages will be removed and replaced with some others like ` libboost-python3-dev libboost-numpy3-dev ` built for python**3**.

At first, when I PR'ed pyarpack, I was thinking that "maybe it's yet too soon to suppress python2 support, even if, I agree this should be done in a near future (I can not say what this near future could be...)"

fghoussen:
@sylvestre : hope this is clear now for you. This is your decision to merge or not this.

sylvestre:
The lib is python3 only in Debian experimental:
https://salsa.debian.org/debian/boost/-/blob/master/debian/control#L1077

The version in Debian testing/unstable is supposed to support both Python 2 and 3:
https://sources.debian.org/src/boost1.67/1.67.0-17/debian/control/#L1054

It is just a matter of time to get only Python 3 support in Debian, therefore I don't see the point of having a Python 2 support.

fghoussen:
> It is just a matter of time to get only Python 3 support in Debian, therefore I don't see the point of having a Python 2 support.

Only a matter of time : yes, I agree.
Don't see the point having python2 support : yes, I agree on "long" term, for "shorter" term I let you decide.

It's all OK / fine to me. The important thing is that you know what's going on : I'am sure you know :D

fghoussen:
BTW, note that pyarpack _should_ be OK on **I**LP64 (this is why you'll find `@PYINT@` in `*.py.in` files), but, this can not be tested on CI: it's only enabled by cmake, no way to get CI to work with cmake on **I**LP64  https://github.com/opencollab/arpack-ng/pull/244#issuecomment-589998647.

fghoussen:
so seems that python2 is dead !...

Issue #232, opened 2019-12-20
CMake with other projects and lib64


thenoursehorse:
## Expected behavior

When linking arpack-ng libraries for other projects using CMake using `-D arpack-ng_DIR=path/to/arpack-ng/lib/cmake`, the variable `arpack_ng_LIBRARIES` should point to the arpack-ng libraries.

## Actual behavior

Sometimes arpack-ng puts libraries in `/path/to/arpack-ng/lib64/` and the `arpack-ng-config.cmake` will not find them correctly because it only searches in `/path/to/arpack-ng/lib/`.

## Where/how to reproduce the problem

- arpack-ng: 3.7.0
- OS: CentOS Linux 7 (Core) x86-64
- compiler: gcc 7.2
- cmake: version 3.9.1
- arpack-ng built with
```
cmake -D EXAMPLES=ON -D MPI=ON -D BUILD_SHARED_LIBS=ON ..
make && make test && make install
```
- My project built with
```
cmake -D arpack-ng_DIR=/path/to/arpack-ng/lib/cmake ..
```

fghoussen:
No centos available, can't test : you may test and PR if it works.

For cmake, you may test something that looks like:
```
~\arpack-ng> git diff arpack-ng-config.cmake.in
diff --git a/arpack-ng-config.cmake.in b/arpack-ng-config.cmake.in
index 30c2b4e..946975e 100644
--- a/arpack-ng-config.cmake.in
+++ b/arpack-ng-config.cmake.in
@@ -7,8 +7,18 @@ if(EXISTS "@CMAKE_INSTALL_PREFIX@/lib/libarpack.a")
 elseif(EXISTS "@CMAKE_INSTALL_PREFIX@/lib/libarpack.so")
   set(arpack_ng_LIBRARIES "@CMAKE_INSTALL_PREFIX@/lib/libarpack.so;@LAPACK_LIBRARIES@;@BLAS_LIBRARIES@" CACHE FILEPATH "arpack-ng: libraries" FORCE)
 endif()
+if(EXISTS "@CMAKE_INSTALL_PREFIX@/lib64/libarpack.a")
+  set(arpack_ng_LIBRARIES "@CMAKE_INSTALL_PREFIX@/lib64/libarpack.a;@LAPACK_LIBRARIES@;@BLAS_LIBRARIES@" CACHE FILEPATH "arpack-ng: libraries" FORCE)
[...]

PR #250, opened 2020-02-29
Don't hard-code installation paths

turboencabulator:
This is a better fix for #232 that doesn't rely on path name assumptions, replacing lib/lib64 with CMAKE_INSTALL_LIBDIR, and include with CMAKE_INSTALL_INCLUDEDIR.  This reverts #235.


sylvestre:
thanks!

PR #254, opened 2020-03-01
pyarpack: add CI job on ILP64.

fghoussen:
## Pull request purpose

Enhancement : add CI job on ILP64 for pyarpack.


.travis.yml
```diff
@@ -254,6 +254,16 @@ jobs:
        echo yes | apt-get -y install intel-mkl libmkl-dev                                                            && \
        apt-get -y install build-essential                                                                            && \
        apt-get -y install git gfortran gcc g++ openmpi-bin libopenmpi-dev automake autoconf libtool pkg-config       && \
+       apt-get -y install libeigen3-dev                                                   && \
```
sylvestre: Most of this is duplicated code from line 181
please move that in a script like
https://github.com/opencollab/arpack-ng/blob/master/scripts/travis_fedora.sh
thanks

fghoussen: No way I go into this burden: such a script would need a parameter for almost each line (!?) as all docker runs are different in many lines, and, it should be tested to pass all CI cases (including this one which has problem others have not).
I can contribute here as long as it sticks reasonably easy/convenient for me too.

fghoussen:
_Very_ slow build both on debian and CI : CI freezes (no output), debian fail on following error (no clue why)
```
libtool: link: ar cru .libs/libpyarpack.a  libpyarpack_la-pyarpack.o
ar: `u' modifier ignored since `D' is the default (see `U')
```

Closing this: no autotool-guru, and, not such a big deal anyway.

.travis.yml
sylvestre: I don't think it is a correct statement.
Most of the apt section is the same
Ditto for boost, etc

>  I can contribute here as long as it sticks reasonably easy/convenient for me too.

I understand that but I won't accept patches adding unnecessary technical debt.

fghoussen: I understand your point.

I think this ` travis_ubuntu.sh ` thing is a (tricky-to-burden) job on his own and should be done in a dedicated PR. This PR was _not_ meant to this: it was meant to offer ` --enable-python3 ` at autotools side for 3 reasons i) to test the python feature on **I**LP64 (were it's **not** tested yet but _should_ work) ii) seems impossible to get cmake to run on **I**LP64 CI https://github.com/opencollab/arpack-ng/pull/244#issuecomment-589998647 iii) offer python from ` configure ` too

fghoussen:
BTW, this does work
```
ar: `u' modifier ignored since `D' is the default (see `U')
```
seems to be only a warning... But build is _super crazy_ slow : libtool can't take templates. cmake does not have this problem.

Conclusion: autotool takes so long to build that ` --enable-python3 ` is likely not an option to have !?...

PR #255, opened 2020-03-02
Rename the python files

sylvestre:
WIP to evaluate with the CI


fghoussen:
Note that in all ` pyarpackSparse*.py.in ` you have ` @PYINT ` to handle LP64 or **I**LP64 :  ` configure_file ` replaces ` PYINT ` by ` int32 ` or ` int64 `  if you run on  LP64 or **I**LP64

```
>> more EXAMPLES/PYARPACK/pyarpackSparseBiCGDiag.py.in
n = 4
i = np.array([], dtype='@PYINT@')
j = np.array([], dtype='@PYINT@')
Aij = np.array([], dtype='float64')

```

sylvestre:
I didn't meant to merge this :'(

fghoussen:
OK :) If I were you I would `rebase -i` master's HEAD to delete d153dff ans addf1b0 (clean history).

Know that the only pyarpack untested feature is : running on **I**LP64, that is, setting ` PYINT="int64" ` : this was the purpose of #254 (this _should_ work as it works on **L**P64 with ` PYINT="int32" `)

The Boost.Python binding **needs** data to be casted in the **exact expected** type (int32, int64, float, double, ...) at python side : without that C++ does **not** get the data the way it expects them and can **not** read python data (AFAIU, python "secretly" stores data "his own way" unless you explicitly cast them the way **you impose** on `np.append`). Tried to mention that here:

- https://github.com/opencollab/arpack-ng/blob/5c988b0f412f3240820cb6f5b3be94af770d5db8/EXAMPLES/PYARPACK/README#L40

- debug traces to help pyarpack if he forgot to cast his data on `np.append` : user https://github.com/opencollab/arpack-ng/blob/5c988b0f412f3240820cb6f5b3be94af770d5db8/EXAMPLES/PYARPACK/pyarpackServices.hpp#L113

- and on all examples, you have cast on `np.append` with explicit comments : https://github.com/opencollab/arpack-ng/blob/5c988b0f412f3240820cb6f5b3be94af770d5db8/EXAMPLES/PYARPACK/pyarpackDenseLDLT.py.in#L13  

Maybe the doc is too skinny... If you think so I could try to get if thicker !... But for people who understand what they do, the API and (commented !) examples should be natural and simple to use.

PR #257, opened 2020-03-03
pyarpack: improve documentation.

fghoussen:
## Pull request purpose

Doc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -578,6 +578,7 @@ if(ICB)
 
   if (PYTHON3)
     python_add_module(pyarpack ${arpackutil_STAT_SRCS} ${arpacksrc_STAT_SRCS} ${arpacksrc_ICB} ${PROJECT_SOURCE_DIR}/EXAMPLES/PYARPACK/pyarpack.cpp)
+    target_compile_definitions(pyarpack PRIVATE PY_MAJOR_VERSION="3")
     set(pyarpack_HDR ${PROJECT_SOURCE_DIR}/ICB ${PROJECT_SOURCE_DIR}/EXAMPLES/MATRIX_MARKET ${PROJECT_SOURCE_DIR}/EXAMPLES/PYARPACK)
     target_include_directories(pyarpack PUBLIC ${pyarpack_HDR} ${EIGEN3_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PYTHON_INCLUDE_DIRS})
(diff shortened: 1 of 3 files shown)


fghoussen:
Trying to ship doc in python module too. Works with python2 here:

```
>>> help(pyarpack)
Help on package pyarpack:

NAME
    pyarpack

DESCRIPTION
    You can use sparse or dense matrices, and, play with iterative or direct mode solvers (CG, LU, ...):
    1. choose arpack solver with a given mode solver
       1.1. if you need to handle sparse matrices
            >> from pyarpack import sparseBiCG as pyarpackSlv
       1.2. if you need to handle dense matrices
            >> from pyarpack import denseBiCG as pyarpackSlv
    2. choose arpack data type (float, double, ...)
       >> arpackSlv = pyarpackSlv.double()
    3. solve the eigen problem
       >> arpackSlv.solve(A, B)
    4. get eigen values and vectors
       >> print(arpackSlv.vec)
       >> print(arpackSlv.val)

    Notes:
    1. arpack data type (float, double, ...) must be consistent with A/B numpy dtypes (float32, float64, ...).
       at python side, the data MUST be casted in the EXACT expected type (int32, int64, float, double, ...).
       otherwise, C++ may not get the data the way it expects them: C++ will not know how to read python data.
       if you are not sure how data have been passed from python to C++, set arpackSlv.debug = 1 and check out debug traces.
       in other words, pyarpack users MUST :
       1.1. create numpy arrays specifying explicitly the type:
            >> Aij = np.array([], dtype='complex128')
       1.2. filling numpy arrays casting value on append:
            >> Aij = np.append(Aij, np.complex128(np.complex( 200.,  200.))) # Casting value on append is MANDATORY or C++ won't get the expected type.
       1.3. calling the solver flavor which is consistent with the numpy array data type:
            >> arpackSlv = pyarpackSlv.complexDouble() # Caution: complexDouble <=> np.array(..., dtype='complex128')
       note: NO data type check can be done at C++ side, the pyarpack user MUST insure data consistency.
    2. sparse matrices must be provided in coo format (n, i, j, Mij), that is, as a tuple where:
       2.1. n is an integer.
       2.2. i, j, Mij are 1 x nnz numpy arrays.
    3. dense  matrices must be provided in raw format (Mij, rowOrdered), that is, as a tuple where:
       3.1. Mij is an n x n numpy array.
       3.2. rowOrdered is a boolean (column ordered if False).
    4. arpack mode solver are provided by eigen:
       4.1. when solver is iterative, A and B can be sparse only.
       4.2. when solver is direct,    A and B can be sparse or dense.


```

PR #260, opened 2020-05-31
[CLEAN - TravisCI]: use bionic as it's now provided by TravisCI !

fghoussen:
## Pull request purpose

Fixing CI.
To test interface64, we need **I**LP64 blas/lapack : the only lib (I know available through CI) which provide them is intel-mkl. intel-mkl is available only in latest debian (but not in "old" ubuntu boxes last time I checked) : need docker+debian => apt-get update / upgrade may change packages !...

Don't remember why `apt-utils` was needed, but, if I added it at the time, there was a reason ! :D Hope it's not needed anymore (would make this a fix quick)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -226,7 +226,7 @@ jobs:
       head -n 50 coveralls.log && tail -n 50 coveralls.log                         \
       &&                                                                           \
       head -n 50 coveralls.json && tail -n 50 coveralls.json
-  # xenial <=> interface64: "recent" systems with ICB + MKL + ILP64 (need debian/testing to get MKL-ILP64)
+  # xenial <=> interface64: "recent" systems with ICB + MKL + ILP64
   #   note: when you PR, docker-cp provides, in the container, the branch associated with the PR (not master where there's nothing new)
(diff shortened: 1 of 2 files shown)


fghoussen:
CI is OK back : merge it first !

fghoussen:
Trying to use bionic as mkl seems now available: likely more stable than debian/testing !

fghoussen:
bionic is now available in TravisCI and seems to be recent enough to provide `mkl` !
Docker could be dropped for all bionic jobs, but, still need to be used for fedora : if OK, would likely be more simple and stable.

fghoussen:
@sylvestre : the last untested feature (_should_ work) will be **I**LP64 + python, if OK up to here, want to try adding it ? (python is available only from cmake, not autotools)

fghoussen:
Bored. From my debian/testing, I _can_ get `mkl` from `docker run ubuntu:bionic` but the exact same thing does not work from CI : bored, return to the way this was done before. Should work.

fghoussen:
Looks OK. Go merge. No better.

fghoussen:
pity, should have fixed up on rebase !... or rebase / push-force. anyway

PR #262, opened 2020-06-02
Eigen version

fghoussen:
## Pull request purpose

Switch eigen version to latest one : https://github.com/opencollab/arpack-ng/pull/259#issuecomment-636797682. On top of #260 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -226,7 +226,7 @@ jobs:
       head -n 50 coveralls.log && tail -n 50 coveralls.log                         \
       &&                                                                           \
       head -n 50 coveralls.json && tail -n 50 coveralls.json
-  # xenial <=> interface64: "recent" systems with ICB + MKL + ILP64 (need debian/testing to get MKL-ILP64)
+  # xenial <=> interface64: "recent" systems with ICB + MKL + ILP64
   #   note: when you PR, docker-cp provides, in the container, the branch associated with the PR (not master where there's nothing new)
   #         1. docker create --name mobydick IMAGE CMD        <=> create a container (= instance of image) but container is NOT yet started
   #         2. docker cp -a ${TRAVIS_BUILD_DIR} mobydick:/tmp <=> copy git repository (CI worker, checkout-ed on PR branch) into the container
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -40,6 +40,14 @@ stages:
 
 jobs:
   include:
+  # fedora (released fedora with openmpi)
```
sylvestre, line 43: fedora / bionic 
please add a comment explaining why we mix the two :)

fghoussen: no idea why ! not me !

fghoussen:
CI can' take it.
> Get:37 http://archive.ubuntu.com/ubuntu xenial/universe amd64 libeigen3-dev all 3.3~beta1-2 [663 kB]
> ...
> checking for EIGEN3... no
> configure: error: Eigen not found. Give hints with PKG_CONFIG_PATH. Disable the need of this dependency using --disable-icb-exmm

@turboencabulator : I guess this is because CI provides "only 3.3.beta ? Do you kow how to overpass that ? If no, I'll close this PR.

.travis.yml
dschwoerer: Not sure what the question is?
Why fedora? Why openmpi? Why released?

I added an openmpi and an mpich case, to test both mpi versions, and also an released and still-in-development version of fedora, also to increase coverage. Why I mixed the on with the other, no real reason.

fghoussen: Not sure what the question was either ! Good for me.

fghoussen:
No idea with this fails. Seems unrelated with eigen version...

fghoussen:
WTF ?!...

> Scanning dependencies of target arpackmm
830[ 64%] Building CXX object CMakeFiles/arpackmm.dir/EXAMPLES/MATRIX_MARKET/arpackmm.cpp.o
831virtual memory exhausted: Cannot allocate memory

fghoussen:
Eigen 3.3. is available with bionic but trigger "831virtual memory exhausted: Cannot allocate memory" error. Eigen 3.3 is not available with xenial : cmake/autotools break, but, CI is OK.

fghoussen:
@sylvestre : only solution I see for now, suppress `exmm` option for coverage in CI. May lower CI a (just) bit (`exmm` => Eigen Template => need lot of memory to compile + longer compile time) : your decision

fghoussen:
@sylvestre : pass if you accept a small coverage decrease. Your decision !

dschwoerer:
Another option to decrease memory consumption is decrease debugging level. You could try to use `-g1` rather then `-g`.

sylvestre:
please update the changelog when you have a chance :)

PR #265, opened 2020-06-05
TravisCI: add centos build.

fghoussen:
## Pull request purpose

centos build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ services:
 stages:
   # order stages
   - name: opensuse
+  - name: centos
   - name: fedora
   - name: osx
   - name: ubuntu_precise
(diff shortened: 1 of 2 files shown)


fghoussen:
got similar pb on centos. no idea why ?! @dschwoerer : any clue ?

dschwoerer:
It seems you are missing the *-devel files, namely `openmpi-devel`, `blas-devel`, `lapack-devel`

fghoussen:
already tried that... but this was before upgrade

dschwoerer:
Just wanted to note some additional repositories might be required, for centos 8: `dnf config-manager --set-enabled PowerTools` is required for epel.
https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F

fghoussen:
not used to centos. trying tricks provided by collegues. will give up if dead end.

fghoussen:
@dschwoerer : thanks !

sylvestre:
Excellent, you could probably add this to the release notes?
CI: Support for centos added in automation

fghoussen:
I feel always using same compilers`gcc` and `openmpi` which breaks here... Try to add another build with `autotools` + `clang` ?...

fghoussen:
`openmpi` sucks on centos... no better than this

dschwoerer:
What issues are you having with openmpi on centos?

fghoussen:
@dschwoerer : tests crash. don't know why. added error log tail in hope to get traces

fghoussen:
> mpirun has detected an attempt to run as root. Running as root is *strongly* discouraged as any mistake (e.g., in defining TMPDIR) or bug can result in catastrophic damage to the OS file system, leaving your system in an unusable state. We strongly suggest that you run mpirun as a non-root user. You can override this protection by adding the --allow-run-as-root option

Oh men ! Already hit this problem : `openmpi` crashes because it's doesn't like been ran by root and you must be root in docker. `openmpi` could be used with that kind of stuff https://github.com/opencollab/arpack-ng/blob/e50818137a1996134df9105841ca099011674b8c/.travis.yml#L214 on `CMakeLists.txt`

dschwoerer:
That is however not the only problem, see #268 for more issues.
https://travis-ci.org/github/opencollab/arpack-ng/jobs/697378825

Unfortunately I don't know what causes this.

fghoussen:
May be related : https://bugzilla.redhat.com/show_bug.cgi?id=1744780, https://github.com/open-mpi/ompi/issues/6671

fghoussen:
BTW, as `openmpi` is broken, reset-hard to `mpich`.

Issue #252, opened 2020-03-01
configure isn't Bourne shell compatible


yurivict:
```
checking whether mpic++ supports C++11 features by default... yes
./configure: ${EIGEN3_CFLAGS/...}: Bad substitution
```

According to GNU standards, ```configure``` should be in Bourne shell. https://www.gnu.org/software/autoconf/manual/autoconf-2.64/html_node/Portable-Shell.html

fghoussen:
@yurivict : does #253 help ? @turboencabulator : you tagged FIXME, what do you think ?

yurivict:
@fghoussen Yes, this solved the problem.

turboencabulator:
@fghoussen I'm queuing up some changes to get rid of most of these FIXMEs.  Seems good to me for the meantime.

PR #267, opened 2020-06-06
TravisCI: add freebsd-like build.

fghoussen:
## Pull request purpose

Just realizing that docker may also emulate `sh` (not `bash` !) : freebsd is not available but one big feature is `sh` (#252)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ services:
 
 stages:
   # order stages
+  - name: sh
   - name: fedora
   - name: osx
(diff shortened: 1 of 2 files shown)


fghoussen:
@sylvestre: not sure you want to keep that ? In case yes, `tcsh` (the other freebsd shell) maybe (?) added

sylvestre:
Bof, not sure what is the point :)

fghoussen:
OK :D Closing so !

Issue #128, opened 2018-06-23
icb_parpack_cpp test is failing on ppc64el


sylvestre:
https://buildd.debian.org/status/fetch.php?pkg=arpack&arch=ppc64el&ver=3.6.1-1&stamp=1529756139&raw=0

```
rank 0 - 1000.15 1000.15
Correct eigenvalues not computed
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
FAIL icb_parpack_cpp (exit status: 1)
```

sylvestre:
@fghoussen Does it ring a bell?
we could just disable mpi on this arch...

fghoussen:
Not really: soundless bell ! In build dependencies, I see libopenmpi**3**: could you try with openmpi**2** ?

fghoussen:
If switching from openmpi2 to openmpi3 is OK on ppc64el (?), try it also on armel

fghoussen:
If libopenmpi3 is not to blame (?), then C++11 could be to blame....

 I've just seen in the (armel) log (at the end) :
```
libtool: link: mpic++ -std=gnu++11 ...
E: Build killed with signal TERM after 150 minutes of inactivity

```
The build seems to hang. Looks like the compiler goes crazy because he doesn't know what to do with the (hard-coded) gnu++11 ?!... Really not sure (what is the compiler behind ?). I have no good answer here. I PR a few lines: kill them if they don't go the way you like

10110111:
Looks more like a test hanging than the compiler disliking `gnu++11`. The compiler seems to have finished by the time the timeout message is printed.

fghoussen:
Yeah maybe. Can you have more verbose logs ? Or compile with ` make VERBOSE=1` ?

10110111:
I guess not. But from the log it seems that the first test in the MPI directory hangs. I suppose only adding some prints to the test would help debugging.

dbeurle:
I'm not entirely sure about this, but in the CMake it links against the CXX and the C libraries for MPI.  Is this likely a problem?

Other possibility is a dead lock :/

fghoussen:
[...]

PR #271, opened 2020-06-20
TravisCI: add ppc64le build.

fghoussen:
## Pull request purpose

Related to #128.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ services:
 
 stages:
   # order stages
+  - name: ppc64le
   - name: fedora
   - name: osx
   - name: precise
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -40,6 +41,37 @@ stages:
...
+  # ppc64le: "recent" systems with ICB
+  #   note: when you PR, docker-cp provides, in the container, the branch associated with the PR (not master where there's nothing new)
+  #         1. docker create --name mobydick IMAGE CMD        <=> create a container (= instance of image) but container is NOT yet started
+  #         2. docker cp -a ${TRAVIS_BUILD_DIR} mobydick:/tmp <=> copy git repository (CI worker, checkout-ed on PR branch) into the container
+  #                                                               note: docker-cp works only if copy from/to containers (not images)
+  #         3. docker start -a mobydick                       <=> start to run the container (initialized with docker-cp)
+  - stage: ppc64le
+    dist: bionic
+    script: |
+      sudo docker pull ppc64le/debian                                                                                   \
```
sylvestre, line 53: Please move all of that it to a script which can take arguments just like it is done for redhat

Also, not sure it should be first in the list :)

fghoussen:
KO ! Likely x64 image and needed ppc Image ?

> standard_init_linux.go:190: exec user process caused "exec format error"

PR #278, opened 2020-07-25
[TravisCI] travis_debian.sh: test BP/ILP64 on debian-based arch.

fghoussen:
## Pull request purpose

[TravisCI] Integrate ILP64 in travis_ubuntu.sh.

PR #279, opened 2020-07-27
[BUG FIX] coverage.

fghoussen:
With newer versions of cmake, an unquoted 6th argument gives the
following error:

    STRING sub-command REGEX, mode REPLACE needs at least 6 arguments
    total to command.

Quoting this fixes the issue.

## Pull request purpose

fixing coverage : same problem than #278 in boost.
coverage is often wrong (results changing when commiting comments or CMake/autotools files !...) and not so stable : this could be the reason why... May get coverage more stable and even increase it.


sylvestre:
thanks

PR #280, opened 2020-08-04
[TravisCI] OS X: running the whole test suite.

fghoussen:
## Pull request purpose

Was told this is supposed to work ?!.... checking... Related to #6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,30 +82,7 @@ jobs:
     name: "Fedora rawhide with mpich"
     dist: bionic
     script: ./scripts/travis_fedora.sh setup mpich rawhide
-
-  # osx <=> macOS 10.13 (high sierra)
-  - stage: osx
-    os: osx
-    osx_image: xcode11
-    before_install:
-    - brew list -1 | while read line; do brew unlink $line; done;
-    - brew list gcc      || brew install gcc
-    - brew list cmake    || brew install cmake
-    - brew list openblas || brew install openblas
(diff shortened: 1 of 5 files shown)


fghoussen:
Nope !...

> FAIL icb_arpack_cpp
> ERROR: iparam[4] 4, nev 9, info -9999
> libc++abi.dylib: terminating with uncaught exception of type std::domain_error: Error inside ARPACK routines

fghoussen:
@sylvestre : closing ? or want to keep partial build/test with veclib ?

fghoussen:
@sylvestre : seems `-ff2c` make all tests pass... not sure you want to keep that kind of "switch hand-tuning"

fghoussen:
Seems that with `-ff2c`, all tests pass on mac. Should break with MPI (at least with autotools)...

fghoussen:
Conclusion: full build/test OK if `-ff2c` but **without** MPI. Adding MPI breaks build/test : no idea why ?!...

fghoussen:
Unexpectedly working ?!... :D

fghoussen:
Not sure there's more simple : I guess "switch hand-tuning" is needed on mac (not sure this could be handled by build systems in an "automatic" way)

fghoussen:
@sylvestre: you could make build systems add flags under a "if mac", not sure this is the best thing to do... This is your decision

sylvestre:
Could you please update the changelog to mention that we have now mac support in the CI?

PR #298, opened 2021-03-12
add C99 standard checking to `CMakeLists.txt` & minimum CMake version changed to CMake3

RyanBernX:
older compilers may not enable this by default.

## Pull request purpose

- fix the comment in https://github.com/opencollab/arpack-ng/issues/297
- `add_library( ... INTERFACE ...)` is not supported in cmake 2

PR #299, opened 2021-03-15
Fix ci

RyanBernX:
## Pull request purpose

fix travis ci

- Add `--security-opt seccomp:unconfined` to stage: opensuse as suggested in https://www.reddit.com/r/openSUSE/comments/lm48zo/install_git_in_docker_tumbleweed/
- `centos-stream` renames `PowerTools` to `powertools`
- `brew link` only works for formulae, thus add `--formula` to `brew list`
- ubuntu eoan has reached EOL. The updates should be installed from  http://old-releases.ubuntu.com/ubuntu/
- Add build for ubuntu groovy
- A bug in cmake causes the build to fail on osx. Work around this issue by modifying `FindMPI.cmake`.
- Remove `-ff2c` `-fno_second_underscore` when compiling with generic `lapack` or `openblas`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,15 @@ addons:
       - diffutils
       - findutils
       - libeigen3-dev
+  homebrew:
+    packages:
+      - gcc@10
+      - autoconf
+      - automake
+      - cmake
+      - lapack
+      - mpich
+    update: false
 
 services:
   - docker
@@ -32,8 +41,6 @@ before_script:
 
 stages:
   # order stages
-  - name: ubuntu_eoan
-  - name: ubuntu_groovy
   - name: osx
   - name: opensuse
   - name: centos
@@ -42,7 +49,9 @@ stages:
   - name: ubuntu_trusty
   - name: ubuntu_xenial
   - name: ubuntu_bionic
+  - name: ubuntu_eoan
   - name: ubuntu_focal
+  - name: ubuntu_groovy
   - name: coverage
   - name: debian_interface64
 
@@ -132,19 +141,16 @@ jobs:
   - stage: osx
     os: osx
     osx_image: xcode12
-    before_install:
-    - brew list --formula -1 | while read line; do brew unlink $line; done;
-    - brew list gcc      || brew install gcc@10
-    - brew list cmake    || brew install cmake
-    - brew list lapack   || brew install lapack
-    - brew list mpich    || brew install mpich
-    - brew list --formula -1 | while read line; do brew link --overwrite $line; done;
+    before_script:
     - export CC=gcc-10
     - export CXX=g++-10
     - mpicc -show
+    # manually fix a bug in cmake + osx
+    # this line should be removed when cmake officially fixes it
     - sed -i '' 's/pi\[ce\]/pi[ce]|ramework/' $(brew --prefix cmake)/share/cmake/Modules/FindMPI.cmake
-    # full build and testing with openblas and lapack
-    # note: -ff2c to convert fortran code to C, -fno-second-underscore to force underscoring of external symbols to link
+    # full build and testing with blas and lapack
+    # note: -ff2c and -fno-second-underscore are NOT needed when using generic
+    # blas/lapack
     script: |
       mkdir -p build && cd build && cmake -DBLA_VENDOR=Generic -DCMAKE_PREFIX_PATH=$(brew --prefix lapack) -DEXAMPLES=ON -DICB=ON -DMPI=ON .. && make VERBOSE=1 && make test
   #- stage: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -151,7 +151,8 @@ jobs:
     before_install:
     - brew list -1 | while read line; do brew unlink $line; done;
     - brew list arpack   || brew install arpack # Test osx formula
-    - brew list -1 | while read line; do brew link --overwrite $line; done;
+    # note: brew has updated itself so we must add --formula to only list formulae
+    - brew list --formula -1 | while read line; do brew link --overwrite $line; done;
     script: ls # Need fake script.
   - stage: osx
     os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,13 +136,12 @@ jobs:
     - brew list --formula -1 | while read line; do brew link --overwrite $line; done;
     - export CC=gcc-10
     - export CXX=g++-10
-    - export DYLD_LIBRARY_PATH=$(brew --prefix lapack):${DYLD_LIBRARY_PATH}
     - mpicc -show
     - sed -i '' 's/pi\[ce\]/pi[ce]|ramework/' $(brew --prefix cmake)/share/cmake/Modules/FindMPI.cmake
     # full build and testing with openblas and lapack
     # note: -ff2c to convert fortran code to C, -fno-second-underscore to force underscoring of external symbols to link
     script: |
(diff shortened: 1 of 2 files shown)


fghoussen:
centos is dead (https://arstechnica.com/gadgets/2020/12/centos-linux-is-gone-but-its-refugees-have-alternatives) so we may drop the CI job ? ubuntu eoan should have LTS (?) : would be better to keep it (to test all ubuntu's from 'old' precise to 'new' focal)  @sylvestre what do you think ? @RyanBernX thanks for help !

RyanBernX:
> centos is dead

Yes and no. On one hand, `centos 7` is currently supported until 2024, on the other hand, `centos-stream` serves as an upstream of RHEL. So instead of dropping the CI job, I suggest to add a stage based on `centos 7` (we can drop it in 2024). It is also worth considering to keep the current centos build in order to keep track with RHEL.

> ubuntu eoan should have LTS (?)

I think it doe not. See [this page](https://wiki.ubuntu.com/EoanErmine/ReleaseNotes). We should test the "newer" ubuntu groovy (20.10) instead.

BTW, the CI job failed due to docker's limitation on pulling. Maybe some of you can restart the build?

RyanBernX:
A few updates:
- Build with `centos 7` is added and confirmed to work.
- osx + autoconf + Accelerate.framework is confirmed to work.
- osx + cmake + openblas is still broken. Seems that cmake cannot find mpich although it is installed. Still looking into this ...

fghoussen:
- centos : OK to keep it if it's possible.

- eoan :  better to keep it (to get all distro from precise - precise has LTS why eoan won't), please get it back.

- groovy : OK to add it (now if you like or after - no hury)

- osx  : don't want to cope with cmake bug on CI => could you just replace cmake + openblas + mpich with cmake + blas-or-atlas + mpich-or-openmpi ? => if this very minimal change work, it will be fine enough

- Don't want to mix things up here. First, CI must be fixed .Then #298 will be merged : please remove it from this PR

RyanBernX:
@fghoussen The difference between precise and eoan is we can still install the packages from http://archive.ubuntu.com/ for precise, while the packages of eoan must be installed from http://old-releases.ubuntu.com/ubuntu/ .
I can't tell the exact reason since precise, trusty, eoal all reach EOL.
So a modification of `/etc/apt/sources.list` should be added before `apt install`. I can fix it if necessary.

> Don't want to mix things up here. First, CI must be fixed .

Okay. However, the CI for centos 7 should also be removed since it requires the correct version of `CMakeLists.txt`. It can be included in another PR.

> osx : don't want to cope with cmake bug on CI => could you just replace cmake + openblas + mpich with cmake + blas-or-atlas + mpich-or-openmpi ? => if this very minimal change work, it will be fine enough

Still looking into this. I have no idea how to work around the cmake bug without modifying `FindMPI.cmake`. Anyway, a patched version of `FindMPI.cmake` should work (we should remove it when cmake officially fixes this issue).

scripts/travis_ubuntu.sh
```diff
@@ -3,10 +3,17 @@
 ## -x be verbose; write what we are doing, as we do it
 set -ex
 
+cmd=ls
+if [ $3 = "old" ]; then
```
fghoussen, line 7: Or if no other solution, killing old and replacing with `if [ $2 = "eoan" ]; then ` ? What do you think ?

.travis.yml
```diff
@@ -129,32 +129,38 @@ jobs:
...
-    - brew list -1 | while read line; do brew link --overwrite $line; done;
-    # full build and testing with openblas and lapack
-    # note: -ff2c to convert fortran code to C, -fno-second-underscore to force underscoring of external symbols to link
+    - brew list --formula -1 | while read line; do brew link --overwrite $line; done;
+    - export CC=gcc-10
+    - export CXX=g++-10
+    - mpicc -show
+    # manually fix a bug in cmake + osx
+    # this line should be removed when cmake officially fixes it
+    - sed -i '' 's/pi\[ce\]/pi[ce]|ramework/' $(brew --prefix cmake)/share/cmake/Modules/FindMPI.cmake
```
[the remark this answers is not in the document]
RyanBernX: Ok. I'll give it a try.

scripts/travis_ubuntu.sh
```diff
@@ -3,10 +3,17 @@
...
+    cmd="sed -i 's/\(security\|archive\).ubuntu/old-releases.ubuntu/g' /etc/apt/sources.list"
+fi 
+
```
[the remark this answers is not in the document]
RyanBernX: > If this is removed : build fails ?

Yes. Because `http://archive.ubuntu.com/ubuntu/dists/eoan*` no longer exists, you won't be able to install anything with the original `sources.list`.

scripts/travis_ubuntu.sh
RyanBernX: We may encounter EOLs of many non-LTS ubuntu systems in the future. Thus I use the `old` option to tell the script that this is an EOL system so it will try to subsititute the mirror address. I can remove it if you think `if [ $2 = "eoan" ]` way is better.

.travis.yml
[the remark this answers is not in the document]
RyanBernX: confirmed: osx + openmpi **works**

scripts/travis_ubuntu.sh
RyanBernX: I have no idea how to get ubuntu eoan to work without changing `sources.list`. I've removed the `old` option and replaced with `if [ $2 = ":eoan" ]; then`.

scripts/travis_ubuntu.sh
```diff
@@ -3,10 +3,18 @@
...
+if [ $2 = ":eoan" ]; then
+    cmd="sed -i 's/\(security\|archive\).ubuntu/old-releases.ubuntu/g' /etc/apt/sources.list"
+fi 
+
 sudo docker pull "$1$2"                                                                                           \
 &&                                                                                                                \
 sudo docker create --name mobydick "$1$2" /bin/bash -c                                                            \
-"apt-get    update                                                                                             && \
+"${cmd} &&
+                     \
```
fghoussen, line 16: @RyanBernX : thanks ! Looks great. TravisCI seems to fail to start on last push (?!) : can you try to remove the above extra empty line to see if CI is triggered OK ?

RyanBernX: I've added another commit. Could you please review https://github.com/opencollab/arpack-ng/pull/299/commits/3f66e11946271df3f4420e4a01c6f30e06958f20 ?

scripts/travis_ubuntu.sh
```diff
@@ -3,10 +3,17 @@
...
+cmd=ls
+if [ $2 = ":eoan" ]; then
+    cmd="sed -i 's/\(security\|archive\).ubuntu/old-releases.ubuntu/g' /etc/apt/sources.list"
+fi 
+
 sudo docker pull "$1$2"                                                                                           \
 &&                                                                                                                \
 sudo docker create --name mobydick "$1$2" /bin/bash -c                                                            \
-"apt-get    update                                                                                             && \
+"${cmd} &&
```
fghoussen, line 15: @RyanBernX : before merging, can you add a `\` at the end of line ? Not sure how this works without it. Thanks for your help !

RyanBernX: @fghoussen Added. BTW I'll skip this CI since https://github.com/opencollab/arpack-ng/pull/299/commits/3f66e11946271df3f4420e4a01c6f30e06958f20 has been confirmed to work.

fghoussen:
@RyanBernX : thanks !

PR #303, opened 2021-03-27
[TravisCI] Add windows job.

fghoussen:
## Pull request purpose

Don't merge : works OK on W10 laptop, KO on CI (`makepkg` does nothing).
Let it here in case someone can help... @RyanBernX @eugene523 @ChrisNabold : any clue ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ jobs:
   # windows
   - stage: windows
     os: windows
-    name: "windows with cmake"
+    name: "windows"
     before_install:
       - choco install -y msys2
       - choco upgrade --no-progress -y msys2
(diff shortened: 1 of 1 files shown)


RyanBernX:
This error message may be useful:
```
System is unknown to cmake, create:

Platform/MINGW64_NT-10.0-17763 to use this system, please post your config file on discourse.cmake.org so it can be added to cmake
```

Seems that MINGW64 isn't recognized by cmake therefore `cmake` refuses to add the dependency (for windows).
However, `cmake` **does** recognize MSYS2 platform. Perhaps you can start a pure msys2 shell and set `PATH` for mingw64?

fghoussen:
At least 2 problems here:

- First, lapack is not built on CI : `makepkg` does nothing. May be because the CI box is "old" (W7 or before ?). OK on W10 but building lapack takes long : likely to hit time limit (CI job = 40 or 50 minutes max)

- Second, seems that MINGW64 isn't recognized by cmake : rebased, let's see if 51b299cba0cde3c27817309eb48a3d7f890967e9 makes a difference

> Perhaps you can start a pure msys2 shell and set PATH for mingw64?

My understanding is that this is exactly what this is trying to achieve ! Not sure. May be wrong (not windows based)

RyanBernX:
@fghoussen  May found the problem.

1) Line 507-508
```
The system cannot find the path specified.
Cannot set specified working diretory "MINGW-packagesmingw-w64-lapack".
```
The backslash in `-where MINGW-packages\mingw-w64-lapack` seems gone. Try `\\` or `/` ?

BTW, we may as well install `mingw-w64-x86_64-openblas` instead of compiling blas/lapack.

2) According to [this page](https://discourse.cmake.org/t/mingw-w64-gcc-g-platform/1602), we have to install `mingw-w64-x86_64-cmake` in the MINGW64 shell, instead of `cmake`. Therefore, in the `before_install` section, it should be
```
pacman -Sy --noconfirm mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake make git # gcc, gcc-fortran are not needed
```

fghoussen:
On W10 laptop I tested on, `gcc gcc-fortran` were needed, `openblas` was not available, and I succeeded to build `lapack` from 
`MINGW-packages` (long build > 40 minutes). Trying `openblas` (no build, easy, is it available ?) as next step... And see what happens !

RyanBernX:
In my test environment, `mingw-w64-x86_64-toolchain` already contains `gcc gfortran` built for **mingw64** system (see `/mingw64/bin`. If you simply install `gcc gcc-fortran` (i.e. without the prefix), you will end up with a **msys2** version of gcc/gfortran (see `/usr/bin`). Since we are triggering a **mingw64** build, the platform of the toolchains should match (except `git`, which has nothing to do with build).

Similarly, there is a pre-built openblas for mingw64, named `mingw-w64-x86_64-openblas` (do not simply drop the prefix, otherwise it won't work). I suggest building with this.

In summary:
- Start msys2 and activate the mingw64 environment
- Install: `mingw-w64-x86_64-toolchain` (contains gcc/gfortran/make for mingw64)  `mingw-w64-x86_64-cmake` `make` `git`
- Configure: `cmake -G "MSYS Makefiles" ..` (we may as well manually specify the compilers)
- Make: `make` (note: this `make` comes from msys2, not mingw64)
- Test: `export PATH=$(pwd):$PATH && make test`

Maybe I am still missing something. Feel free to check whether the above method works.

fghoussen:
Tried a few things here: let's see...

fghoussen:
Moved to github actions : windows build OK, partial testing OK.

PR #304, opened 2021-03-29
CI: add support for centos:7

RyanBernX:
## Pull request purpose
CI: add build for centos7

- See https://github.com/opencollab/arpack-ng/pull/299#issuecomment-801342448

### DIFF
--- a/scripts/travis_centos.sh
+++ b/scripts/travis_centos.sh
@@ -21,27 +21,27 @@ else # for 8+, stream
     cmd_powertools="dnf config-manager --set-enabled powertools &&"
 fi
 
-sudo docker pull "${prefix}centos$1"                                        \
-    &&                                                                      \
-    sudo docker create --name mobydick ${prefix}centos$1 /bin/bash -c       \
-    "$pm install -y ${pm}-plugins-core epel-release                      && \
-    $pm upgrade -y                                                       && \
-    $cmd_powertools                                                         \
-    $pm install -y git make gcc gcc-gfortran gcc-c++ environment-modules && \
-    $pm install -y $cmake                                                && \
-    $pm install -y mpich-devel                                           && \
-    $pm --enablerepo=\"epel\" install -y openblas-devel lapack-devel     && \
-    . /etc/profile.d/modules.sh                                          && \
-    module avail && module load mpi && module list                       && \
-    cd /tmp                                                              && \
-    cd arpack-ng                                                         && \
-    git status                                                           && \
-    git log -2                                                           && \
-    mkdir -p build && cd build                                           && \
-    $cmake -DEXAMPLES=ON -DMPI=ON -DICB=ON ..                            && \
-    make all && make test"                                                  \
-    &&                                                                      \
-    sudo docker cp -a ${TRAVIS_BUILD_DIR} mobydick:/tmp                     \
-    &&                                                                      \
-    sudo docker start -a mobydick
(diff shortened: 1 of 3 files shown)


fghoussen:
I didn't follow what happened at centos side : seems killed by IBM, no ?
Why do you think is it a good thing to keep track of centos7 ?

RyanBernX:
**In short**: IBM has sentenced centos to death, but not now. centos 7 will live until 2024.

Currently there are 3 centos-based (supported) distros:
- centos 7 (EOL: 2024)
- centos 8 (EOL: 2021), there will be no centos 9
- centos-stream 8 (EOL: 2024), will evolve to something like centos-stream 9

[This link](https://centos.org/distro-faq/) discusses about the future of centos. Q11 and Q12 are related to CI.
IMO, The CentOS team suggests to use RHEL platforms for testing. If not available, choose centos-stream containers instead.
Meanwhile, there are many centos 7 (RHEL7) based production environments (E.g. HPCs). ARPACK can be installed on those platforms.

The main difference between centos 7 and centos-stream is: centos 7 is the **downstream** of RHEL7, while centos-stream is the **upstream** of RHEL. Therefore, tests on centos 7 will directly reflect the compatiblilty of ARPACK on RHEL7, while tests on centos-stream is a bit less reliable (since the centos team encourages the users to use RHEL). That is the reason to keep track of centos 7 (until 2024).

The best option for me:
- Support the build on centos 7 until 2024
- Optionally support the build on centos-stream

What do you think?

fghoussen:
> What do you think?

I think you are right! 
Does it make sense to track centos-stream from now ? Or too soon (centos-stream not yet enough stable) ?  
Could you wrap centos (7, 8 - and maybe centos-stream ?) jobs in a `./scripts/travis_centos.sh` bash script (like fedora / ubuntu with a `$1` argument) : this would make centos jobs management easier.

RyanBernX:
> Does it make sense to track centos-stream from now ? 

We can do that if you are OK with one of the followings:
1) The official centos-stream container image can be obtained from [this page](https://quay.io/repository/centos/centos?tab=tags). However, we have to create a personal RedHat account if we want to pull from this repo (after June 2021).
2) Or it is possible to turn to some unofficial repos (such as [this one](https://hub.docker.com/r/tgagor/centos-stream) ). However, there is no up-to-date guarantee.

> Could you wrap centos (7, 8 - and maybe centos-stream ?)

I'll give it a try.

fghoussen:
Let's keep this simple for now: stick to centos 7 and 8. Does latest == 8 ?

RyanBernX:
> Let's keep this simple for now: stick to centos 7 and 8. Does latest == 8 ?

Yes. Waiting for CI to complete...

scripts/travis_centos.sh
```diff
@@ -0,0 +1,47 @@
...
+    pm="yum"
+    cmake="cmake3"
+else # for 8+, stream
+    pm="dnf"
+    cmake="cmake"
+    cmd_powertools="dnf config-manager --set-enabled powertools &&"
+fi
+
+sudo docker pull "${prefix}centos$1"                                        \
+    &&                                                                      \
```
fghoussen, line 25: Can you reindent from here: would make script more readable. Thanks.

RyanBernX: Done.

sylvestre:
image

made my day :)

Issue #309, opened 2021-06-05
All docker runs fail


fghoussen:
## Expected behavior

Docker runs should be OK.

## Actual behavior

Docker runs fail with this kind of errors
> +sudo docker pull centos:latest
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

## Notes, remarks

Not sure what the process should be, but, my understanding from https://github.com/marketplace/actions/docker-login (github container registry GCR) is that :

1. @sylvestre , you must (looks like I can't) enable GCR for the opencollab organization (https://docs.github.com/en/packages/working-with-a-github-packages-registry/enabling-improved-container-support-with-the-container-registry - Enabling the Container registry for your organization account)
2. push this
```
>> git diff
diff --git a/.travis.yml b/.travis.yml
index 9cdfda8..5084edd 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,14 @@ stages:
   - name: debian_interface64
 
 jobs:
+  login:
+    steps:
+      - name: Login to GitHub Container Registry
+      - uses: docker/login-action@v1.9.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.opencollab }}
+          password: ${{ secrets.GITHUB_TOKEN }}
   include:
   # opensuse: "recent" systems with ICB
   #   note: when you PR, docker-cp provides, in the container, the branch associated with the PR (not master where there's nothing new)
```

... And hope this is enough to get docker back ?!...

fghoussen:
@sylvestre  : did you had time to have a look at this ?

sylvestre:
@fghoussen i missed your message, sorry
I will have a look asap

fghoussen:
@sylvestre : as far as I understand (not sure), you may try, as github/opencollab **admin** (I'm not), this 3-steps process

1.  Create a docker account here https://hub.docker.com/signup and a personal access token (PAT) for docker https://docs.docker.com/docker-hub/access-tokens

2. Create a secret for github/opencollab (you need to be **admin** for that) where you store your docker login and PAT : https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization

3. Try to push something like this on a test branch... And check out if this works or not

```
>> git diff
diff --git a/.travis.yml b/.travis.yml
index 9cdfda8..5084edd 100644
--- a/.travis.yml
[...]

PR #315, opened 2021-09-05
[CI] Fix docker error: toomanyrequests You have reached your pull rat…

fghoussen:
Try fixing #309

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -130,7 +130,7 @@ jobs:
       - name: Login to Docker Hub
         run: sudo docker login -u "${{ secrets.DOCKER_USERNAME }}" -p "${{ secrets.DOCKER_TOKEN }}"
       - name: Run job
-        run: ./scripts/travis_fedora.sh setup mpich fedora
+        run: ./scripts/travis_fedora.sh setup mpich
   docker_debian:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


fghoussen:
No-op commit ?!...

fghoussen:
@sylvestre: travis CI seems to be back. Let's see if docker runs...

fghoussen:
@sylvestre : do you have a plan for opencollab travis-ci account ?

image

fghoussen:
Try running arm/ppc ? @sylvestre : what do you think ?

fghoussen:
Travis down.
GH : osx / docker OK, old distros (bionic to precise) / coverage (which sounds bad more than old distros) KO.
@sylvestre : is GH a good move ? Or the only one we could take ? What do you think ?

dimpase:
Travis went expensive last year, or when was that, so no, it's not free any more even for open-source (only few runs per month, something like this). GitHub Actions are the right move.

dimpase:
Can we split this PR into two, leaving the travis-related stuff aside?

fghoussen:
> Travis went expensive last year, or when was that, so no, it's not free any more even for open-source. GitHub Actions are the right move.

Didn't know ! Thanks for the tip :D

> Can we split this PR into two, leaving the travis-related stuff aside?

Still didn't decided what to do here.
For me : drop coverage (too bad), drop old distro (precise, ...), drop travis and move on to gh
@sylvestre  : what do you think ?

fghoussen:
@sylvestre : if you don't want to drop coverage, you may try https://github.com/codecov/cpp-11-standard

fghoussen:
Closing as it served as baseline for another merged PR.

PR #317, opened 2021-09-27
CI based on GitHub Actions only

dimpase:
## Travis CI no longer viable

here we replace Travis CI with GitHub Actions CI

-
-


dimpase:
I just ran this here: https://github.com/dimpase/arpack-ng/actions/runs/1279477662

Something wrong with Docker-based runs (the rest are good), but Docker is a bit above my pay grade...

dimpase:
Perhaps we should disable Docker stuff for the time being, after all, it's pretty much orthogonal to the project aim.

fghoussen:
> Perhaps we should disable Docker

No ! Docker is precious : it enables to run other distros (fedora, centos, ...) from anywhere (ubuntu in this case).
Docker works if you set up secrets (what I did for my account and opencollab - that's its OK there but KO at your side) : once merged all PRs should run OK (at opencollab side)

dimpase:
well, then you can go ahead and merge this, right?

sylvestre:
I think the docker solution is pretty heavy and you have now better solution in github actions, for example:
https://github.com/opencollab/llvm-toolchain-integration-test-suite/commit/d3dec7f909531936678d4e6698413bf4c426dd43

sylvestre:
let's try to see if it sticks :)

dimpase:
It worked! https://github.com/opencollab/arpack-ng/actions/runs/1285891897 is all green.

fghoussen:
So we don't have coverage anymore : clean coveralls (cmake files / options) ?

dimpase:
Oops, mea culpa. I didn't notice that you only have

```
on: [push]
```

an not

```
on: [push, pull_request]
```

which prevented triggering Actions for PRs.

I'll fix this later tonight.

dimpase:
see, now on #322 all the GitHub Actions available are triggered.

PR #341, opened 2022-02-06
Fix ci

fghoussen:
## Pull request purpose

fixing CI that seems broken again... help would be appreciated here...
Top priority PR: no CI (arms), no merge (chocolate)...


fghoussen:
CentOS seems _definitely_ dead! @sylvestre : kill? Or try to replace it with RedHat https://faun.pub/how-to-build-rhel-based-docker-images-on-redhat-enterprise-linux-def012f84890?

sylvestre:
yeah, red hat sounds great

fghoussen:
basic redhat podman test: https://access.redhat.com/solutions/3696691

fghoussen:
Are Github actions broken? Commits are not seen on opencollab?!...
https://github.com/opencollab/arpack-ng/actions/runs/1820338627   => opencollab side: centos should be deleted ?!
https://github.com/fghoussen/arpack-ng/actions/runs/1820338492    => my side: redhat test OK

What's going on? Any help/clue is appreciated here

fghoussen:
Brilliant?!... `pull_request` does not see secrets anymore, so docker jobs are down...

fghoussen:
Now #343 is merged... Maybe this PR will test the code modified by this PR?!... Rebasing on master...

fghoussen:
Check commit is OK 

```
commit a40e66c6c6aa68fac4e97c6aca9992dedb1826c4
Author: Franck HOUSSEN <***@users.noreply.github.com>
Date:   Sun Feb 13 13:22:23 2022 +0100

    run travis_redhat.
```

But still does not run `docker_redhat` instead of  `docker_redhat` (which should eb renamed `podman_redhat` BTW!)

fghoussen:
@sylvestre: PR does not run code from PR'ed branch but the one from master?! PR changes are _not_ tested?!

fghoussen:
Same problem: PR is not tested, ea1dfe7916594ec3af4a4f7b6c4afeba51c4a466 (master) is ran, _not_ b5c02b06973818e707eb55f4293b9ae057aedb62 (PR). No solution. any help is welcome @sylvestre ?

fghoussen:
@sylvestre: my understanding (I may be wrong) is that we must drop all docker/podman (redhat) jobs, switch back to `pull_request`... Test the new workflow and hope for it to work as expected...

- To run docker jobs we now need login/password. These login/password are stored as secrets in opencollab user: any user who would PR can not access them, so all docker builds are KO.
- To access them, we need to trigger CI on `pull_request_target` instead of `pull_request`:  this allows any user who would PR to run in 'the env of opencollab master with secrets as env variables' so docker jobs can run... But it seems that the commit is stuck on opencollab master (https://github.com/opencollab/arpack-ng/commit/ea1dfe7916594ec3af4a4f7b6c4afeba51c4a466) but not the one the CI is supposed to test (= https://github.com/opencollab/arpack-ng/commit/b5c02b06973818e707eb55f4293b9ae057aedb62 = commit of the PR => check out `podman_redhat` job : the job ran is the one from opencollab master, not the one the PR intend to modify).

Not sure to know how to get out of this mess... Switching back to `pull_request` will  prevent docker to run, and, I'am not even sure this would make things back as expected (= CI testing the code modified by the PR!)

sylvestre:
not sure what you mean?!
many jobs seems to work
just disable the failing jobs?
it is probably good enough

fghoussen:
Less and less time for that...

As is jobs does _not_ work: https://github.com/opencollab/arpack-ng/pull/341#issuecomment-1054533117

Only solution I see: 1) drop all docker-like jobs (free the workflow from use of secrets - mandatory for docker login) 2) return to `pull_request` (do not use `pull_request_target` that seems to make the workflow based on master) 3) test the resulting  `pull_request` workflow (without docker-like jobs and no need for secrets) 4) make sure the code of a PR is indeed tested by the PR (= PR test the code modified - this is _no_ more what happens!). This means getting back to CI when done on TravisCI with only distro offered by Travis. Sounds to me like the only way. Not sure if there is a better way

@sylvestre: OK ?

sylvestre:
sounds good!

fghoussen:
OK. If you're OK, I'll try to fix the workflow this way when possible (not much time for now).
The workflow is now broken: when one PR, it seems that the code of master is ran (not the one of the PR)... Looks related to the fact that with github actions, all is done to avoid/limit/prevent malicious actions: so simple things are no more simple...

fghoussen:
@sylvestre : drop coverage? Never works with github-actions (but did with TravisCI)

Mac with cmake KO: not mac-based, can somebody help? @prj-: any clue?
>[ 99%] Building C object CMakeFiles/icb_parpack_c.dir/PARPACK/TESTS/MPI/icb_parpack_c.c.o
clang: error: unknown argument: '-fallow-argument-mismatch'

prj-:
Where is `-fallow-argument-mismatch` coming from? It should only be set for the Fortran compiler.

fghoussen:
> Where is `-fallow-argument-mismatch` coming from?

No idea! On github-actions boxes, seems that gcc can not be installed (pkg conflicts): gfortran is used for fortran and clang for C/C++ 
https://github.com/opencollab/arpack-ng/blob/8d62927dec1dc156f880cf2604992a6c2888044e/.github/workflows/jobs.yml#L204

> It should only be set for the Fortran compiler.

I know! Not mac-based and not so familiar with it. This is why I tried f36b044. If you have any clue, would be nice :)

fghoussen:
@sylvestre: are you OK with 0ba37a7?

fghoussen:
OK, my understanding is that problems have been passed for C/C++ with gfortran/clang, but, we hit now same problems for MPI (**p**arpack). 

>[ 99%] Building C object CMakeFiles/icb_parpack_c.dir/PARPACK/TESTS/MPI/icb_parpack_c.c.o
clang: error: unknown argument: '-fallow-argument-mismatch'

PR #342, opened 2022-02-06
[CMake] Test find_package.

fghoussen:
## Pull request purpose

*.cmake, *.pc files can not be tested as it's a difficult task. Result: *.pc/cmake files are blind-pushed.
Ressorting on old commit that never worked https://github.com/opencollab/arpack-ng/pull/259#issuecomment-603187398 : at the time, if I remeber well cmake refused to `find_package` in a *.cmake file... Not sure if this is allowed now

Any help here would be appreciated (@turboencabulator  ?). Related to #333 too.

Would be good to have this before merging PR impacting cmake/pc files...

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -23,8 +23,10 @@ jobs:
           cd build
           cmake -DEXAMPLES=ON -DMPI=ON -DICB=ON ..
           make all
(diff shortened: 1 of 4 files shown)


fghoussen:
CMake doesn't allow to call `add_library` and others from .cmake files... This can not work. No workaround. Dead end.

PR #344, opened 2022-02-13
Fix github actions: checkout branch we need to test (master or PR branch).

fghoussen:
## Pull request purpose

fixing issue : fix github actions ? PR does not checkout PR branch
blind-pushing on master: no choice...


fghoussen:
centos and osx builds are supposed to be and stay broken.

PR #347, opened 2022-04-09
cmake: parse libraries without "lib" prefix.

mmuetzel:
Shared library files might not start with a "lib" prefix. Or a user might pass a library name (without the "lib" prefix). If this happens, the content of the package config files created by cmake might be wrong.

Exemplary steps to reproduce:
```
cmake -DBLAS_LIBRARIES=openblas
```

If configured like this, `arpack.pc` will contain a line like the following:
```
Libs.private: openblas
```
It should be:
```
Libs.private: -lopenblas
```

The proposed change should fix the library name parsing for this case.


CMakeLists.txt
```diff
@@ -751,15 +751,15 @@ set(PACKAGE_URL "https://github.com/opencollab/arpack-ng/")
 set(ARPACK_PC_LIBS_PRIVATE)
```
fghoussen, line 751: ARPACK_PC_LIBS_PRIVATE looks like to be unused in CMakeLists.txt !?... Do you use it? If no, kill it

CMakeLists.txt
```diff
@@ -751,15 +751,15 @@ set(PACKAGE_URL "https://github.com/opencollab/arpack-ng/")
...
 foreach(lib ${LAPACK_LIBRARIES})
     get_filename_component(libname ${lib} NAME)
-    string(REGEX REPLACE "^lib([^.]+).*$" "-l\\1" libname ${libname})
+    string(REGEX REPLACE "^(lib)?([^.]+).*$" "-l\\2" libname ${libname})
     list(APPEND ARPACK_PC_LIBS_PRIVATE "${libname}")
 endforeach()
 string(REPLACE ";" " " ARPACK_PC_LIBS_PRIVATE "${ARPACK_PC_LIBS_PRIVATE}")
 
 set(PARPACK_PC_LIBS_PRIVATE)
```
fghoussen, line 759: PARPACK_PC_LIBS_PRIVATE looks like to be unused in CMakeLists.txt !?... Do you use it? If no, kill it

CMakeLists.txt
mmuetzel: I'm not sure I understand what you mean.
It is used in `SRC/arpack.pc.in`.

CMakeLists.txt
mmuetzel: Same as above.
It is used in `PARPACK/SRC/MPI/parpack.pc.in`.

CMakeLists.txt
fghoussen: OK, not used in CMakeLists.txt but used elsewhere!

CMakeLists.txt
fghoussen: OK, not used in CMakeLists.txt but used elsewhere!

mmuetzel:
The PR got in a strange state where it didn't apply anymore to the master branch. (I don't know how that happened.)
Should be fixed now again. (I hope.)

fghoussen:
I pushed-force master: this is why you had to rebase

sylvestre:
In general, you should not do that ;)

mmuetzel:
> In general, you should not do that ;) Message ID: ***@***.***>

Sorry. I'm confused. Do what? Force push master? Or make the proposed change?

sylvestre:
Force push to master/main
it is too risky as it can remove some history

fghoussen:
> In general, you should not do that ;)

I know! This was very specific: for months, CI is down: I had to blind-push stuffs on master without success to fix CI during weeks. I only pushed-force master regarding these commits only to get a clean history back: changes involve only `.github/workflow/jobs.yml` that nobody needs to change. No history has been lost! In the end, everybody can rebase on top of the new master without any risk (I've done the push-force on master in such a way to ensure that)

mmuetzel:
Would you like some changes to this PR? Is this acceptable like it is?

fghoussen:
Priority was to get CI back. I want to take a chance to integrate other stuffs before: I'll merge you soon.
arpack is dead: maintained-only, no-dev. I try to stick around no more than once a week: no hurry here.

fghoussen:
@mmuetzel: can you update the change log?

mmuetzel:
Ok like this?

PR #355, opened 2022-05-04
cmake: Don't override BLAS/LAPACK linker flags.

mmuetzel:
When OpenBLAS is the only BLAS/LAPACK that is installed on the build system, the static linker flags in the pkg-config files might be wrong.
They contain the following line:
```
Libs.private: BLAS::BLAS
```
It should be:
```
Libs.private: -lopenblas
```

During the configuration step, the following is printed:
```
   -- BLAS:
      -- link:    F:/msys64/ucrt64/lib/libopenblas.dll.a
   -- LAPACK:
      -- link:    BLAS::BLAS
```

That seems to be caused by the following command in the `CMakeLists.txt` file:
```
get_target_property(LAPACK_LIBRARIES LAPACK::LAPACK INTERFACE_LINK_LIBRARIES) # Get variables from target (*.pc/cmake, msg).
```

I'm not sure if this is a cmake bug or if this is the expected result.

Anyway, the value of `LAPACK_LIBRARIES` seems to be set correctly immediately before that command. Simply removing it fixes the content of the `.pc` file.

For consistency, this PR also removes the corresponding command for the `BLAS_LIBRARIES` variable.

See also: https://github.com/msys2/MINGW-packages/issues/11376

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,7 +29,8 @@ arpack-ng - 3.9.0
  * use CMAKE_INSTALL_FULL_<dir> in arpack.pc
 
 [ Markus Mützel ]
- * CMake: handle libraries without "lib" prefix.
+ * CMake: Handle libraries without "lib" prefix.
+ * CMake: Don't override BLAS/LAPACK/MPI flags. Directly use results from the Find* modules instead.
 
  -- Sylvestre Ledru <sylvestre@debian.org> Mon, 07 Dec 2020 11:37:40 +0100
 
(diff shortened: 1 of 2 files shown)


fghoussen:
Thanks for reporting! Sounds relevant.
Can you kill all occurrences of `get_target_property`? If relevant for blas/lapack, same should apply to other dependencies (MPI, eigen, ...), no? Also, change you update the change log?

mmuetzel:
Thanks for the review.

I removed all occurrences of `get_target_property` from `CMakeLists.txt`. That seems to work correctly for me. The content of the generated `.pc` and `.cmake` files look good to me after this change.

I also added a note to the `CHANGES` file.

PR #356, opened 2022-05-07
Test install with ITF64 suffix.

fghoussen:
## Pull request purpose

Install with ITF64 suffix.


fghoussen:
> Do you think this could be standardized somehow?

No idea. I asked myself quite the same question. I do not know what standard recommendations could be. See https://github.com/opencollab/arpack-ng/discussions/358

PR #357, opened 2022-05-16
Avoid isnan(), which is GNU-specific

joseeroman:
The proposed solution relies on IEEE-754 standard definition of NaN,
https://stackoverflow.com/questions/17389958/is-there-a-standard-way-to-check-for-infinite-and-nan-in-fortran-90-95

This solves a problem that appears with `nvfortran` compiler using CMake build. By the way, why is the CMake build compiling the tests? If I am not wrong the `configure` build skips tests (compiles the libraries only).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,8 @@
 arpack-ng - 3.9.0
 
+[ Jose E. Roman ]
+* Avoid using isnan() in tests, since is GNU-specific
+
 [ Tom Payerle ]
 * Change the continuation line format for stat.h, debug.h
 
--- a/TESTS/bug_58_double.f
+++ b/TESTS/bug_58_double.f
@@ -417,7 +417,7 @@ c
          print *, ' '
 c
       end if
-      if (isnan(v(1,1))) then
+      if (v(1,1) /= v(1,1)) then   ! result is NaN
          stop 1
       end if
 c


fghoussen:
Thanks for reporting! Can you update the change log?

> By the way, why is the CMake build compiling the tests? If I am not wrong the configure build skips tests 

AFAIR, both CMake and autotools build/run tests.

joseeroman:
> Thanks for reporting! Can you update the change log?

Done.

> AFAIR, both CMake and autotools build/run tests.

In the autotools build, `make all` (which is the default) does not compile tests, so the `isnan` issue does not arise. You have to `make check` to hit the issue. With CMake, tests are built by default. I tried to disable the tests, but I don't know enough of CMake and my solution worked in some machine but not in other.

fghoussen:
> You have to make check to hit the issue.

This is the expected autotools behavior.
`make check` run tests and compile test code needed only for testing. `make all` compiles core code only.

> With CMake, tests are built by default. I tried to disable the tests.

AFAIR, you can remove examples with `cmake -DEXAMPLES=OFF ..`

joseeroman:
> AFAIR, you can remove examples with cmake -DEXAMPLES=OFF ..

`EXAMPLES` and `TESTS` seem to be different things. It still builds the tests. I tried `-DTESTS=OFF` and it worked on my macOS laptop but not on an Ubuntu server. But anyway, when this PR is merged then building the tests is not a big problem.

PR #360, opened 2022-05-27
Standardized suffix for INTERFACE64

mmuetzel:
This is a proposal for how the suffix could be standardized that is used to indicate that the package was built for 64-bit Fortran integers.

See also #358.


mmuetzel:
I guess this ship has sailed. Different distributions are using different naming conventions already.

PR #361, opened 2022-05-28
Simplify build rules for CI on Windows

mmuetzel:
Use action msys2/setup-msys2 to install the MSYS2 build environment.
Among other things, this caches the downloaded packages which might speed up the installation process. But the main advantage is imho that it simplifies the commands that are used for building, and makes them easier to read.

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -221,30 +221,31 @@ jobs:
           make check
   windows_latest_cmake:
     runs-on: windows-latest
+    defaults:
+      run:
+        # Use MSYS2 as default shell
+        shell: msys2 {0}
     steps:
-      - name: Clone and check out repository code
-        uses: actions/checkout@v2
+      - name: Install MSYS2 build environment
+        uses: msys2/setup-msys2@v2
         with:
-          fetch-depth: 0
-          ref: ${{github.event.pull_request.head.ref}} # Branch where changes are implemented.
-          repository: ${{github.event.pull_request.head.repo.full_name}} # Repo where changes are implemented.
+          update: true
+          msystem: MINGW64
+          install: >-
+            base-devel
+            git
+            mingw-w64-x86_64-cmake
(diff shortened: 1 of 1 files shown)


fghoussen:
Thanks for reporting!

.github/workflows/jobs.yml
```diff
@@ -221,30 +221,31 @@ jobs:
           make check
   windows_latest_cmake:
     runs-on: windows-latest
+    defaults:
+      run:
+        # Use MSYS2 as default shell
+        shell: msys2 {0}
     steps:
-      - name: Clone and check out repository code
```
fghoussen, line 225: Can you set back this whole step? We had the CI broken (c45094c7583485b18d298423a6f5e66a130d3af4) because the branch checkout-ed by the CI job was not the one containing the modified code...

mmuetzel: I don't think that is necessary. Especially, `fetch-depth: 0` is overkill if you don't switch between branches or revisions anyway in any later build step.

mmuetzel: > The "Install MSYS2 build environment" step should be inserted after the "Clone and check out repository code" step.

We need `git` for that step. So, it should be after MSYS2 is installed.

mmuetzel: You might be under the impression that you need to select these specific `repository` and `ref` here. But that is not true.

When you selected the trigger `pull_request_target`, the action checked out the PR *base* branch. But the trigger for this action is `pull_request`. That selects the *merge* branch and commit.
See: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

fghoussen: Totally believe you when you say it may be overkill or not necessary.

But, I prefer to have all jobs shaped the same way: easier to maintain, and, when needed changes are applied to all jobs at once. So for now, I prefer all jobs to be consistent (as it does not hurt)

PR #362, opened 2022-05-28
Create one .cmake file per arpack-ng flavor (32-bits, 64-bits, ILP64).

fghoussen:
## Pull request purpose

Fixing install. Related to #358.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,7 @@ arpack-ng - 3.9.0
 * Support for NAG's nagfor Fortran compiler
 
 [ Franck Houssen ]
+* Create one .cmake file per arpack-ng flavor (32-bits, 64-bits, ILP64).
 * Test autotools pkg-config (*.pc files) with/without LIBSUFFIX/ITF64SUFFIX.
 * Test CMake find_package (*.cmake files) with/without LIBSUFFIX/ITF64SUFFIX.
 * [BUG FIX] autotools: ICB must be checked first (MPI changes compilers).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -806,10 +806,10 @@ configure_file(arpackicb.h.in "${PROJECT_BINARY_DIR}/arpackicb.h" @ONLY)
 install(FILES "${PROJECT_BINARY_DIR}/arpackicb.h" DESTINATION ${ARPACK_INSTALL_INCLUDEDIR})
 
 # Provide find_package for arpack-ng to users.
-configure_file(cmake/arpackng-config.cmake.in "${PROJECT_BINARY_DIR}/arpackng${LIBSUFFIX}${ITF64SUFFIX}-config.cmake" @ONLY)
-install(FILES "${PROJECT_BINARY_DIR}/arpackng${LIBSUFFIX}${ITF64SUFFIX}-config.cmake" DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/arpack-ng) # find_package(arpackng)
-configure_file(cmake/arpackng-config-version.cmake.in "${PROJECT_BINARY_DIR}/arpackng${LIBSUFFIX}${ITF64SUFFIX}-config-version.cmake" @ONLY)
-install(FILES "${PROJECT_BINARY_DIR}/arpackng${LIBSUFFIX}${ITF64SUFFIX}-config-version.cmake" DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/arpack-ng)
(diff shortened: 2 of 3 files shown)


fghoussen:
Resulting in this kind of install dir:

```
/tmp/tstCMakeInstall/local
├── include
│   └── arpack-ngILP64
│       ├── ...
└── lib
    ├── cmake
    │   └── arpack-ngILP64
    │       ├── arpackng-config.cmake
    │       └── arpackng-config-version.cmake
    ├── ...

```

or 

```
/tmp/tstCMakeInstall/local
├── include
│   └── arpack-ng
│       ├── ...
└── lib
    ├── cmake
    │   └── arpack-ng64
    │       ├── arpackng-config.cmake
    │       └── arpackng-config-version.cmake
    ├── ...

```

mmuetzel:
Like I tried to explain in #358, I don't think that this change would be helpful.

fghoussen:
>I don't think that this change would be helpful.

Maybe not for what you proposed initially, but, it improves packaging and make things symmetric (all flavors available - same logic as pc files).

>arpackng-config.cmake must be unchanged as it changes the key given to find_package

You're right: the key given to find_package is unchanged, it's 'arpackng'. To work, `find_package` rely on the env: `export arpackng_DIR=/path/to/cmakefile` enables to select the arpack flavor you need.https://github.com/opencollab/arpack-ng/blob/08e2368179d8a234f64c44d8ff354de17b039c56/cmake/tstCMakeInstall.sh.in#L31

mmuetzel:
> You're right: the key given to find_package is unchanged, it will always be 'arpackng'. To work, `find_package` rely on the env: `export arpackng_DIR=/path/to/cmakefile` enables to select the arpack flavor you need.

I might be missing the point. But isn't one of the main motivations for installing .pc and .cmake files to allow other projects to link to this project without needing to know where and how it is installed. If a prerequisite of using these files is that it is known where they are installed, they are kind of pointless...

fghoussen:
>If a prerequisite of using these files is that it is known where they are installed, they are kind of pointless...

Your remark is 101% relevant! :) I hit the same question years ago: not sure at all, but, at this point, this is how I understand things are supposed to work

1. Say you use `environment-modules`: 

    1. Say you are an admin: you have full rw rights, you download arpack-ng, build it, and install it in a place you know (say `/opt/arpack-ng/local`). Then, you write a module-file in which you `export arpackng_DIR=/path/to/cmakefile`. Indeed, the admin has to know where the root install is (he defined it!) and he has duty to make this information available to the environment (here using `environment-modules`)
 
    2. Now, say you are a (new) user: you just need to `module-load` what you need _without knowing anything_ about how/where the code was compiled and installed. This is where you are happy to use a build system (same idea for autotools and cmake - `configure` rely generally on `PKG_CONFIG_PATH` and `cmake` on `find_package` which rely also on the environment)

2. Say you don't use `environment-modules`: you _need_ to rely on informations that have been set _for you by somebody else_ (finding a package is no magic) and _when_ it's possible you are, here again, happy to use a build system _without knowing anything_ about how/where the code was compiled and installed.

    1. Either `apt-get install intel-mkl` has installed pc files in `/usr/local/lib/pkgconfig` (default location where you may find different pc files for different flavors of the same package) and then the autotools `configure` and/or cmake may rely on it
    
    2. Or kitware as provided _for you_ a FindXXX.cmake in _their own_ files. For instance, on my debian, looking for LAPACK
        
        ```
            >> locate FindLAPACK.cmake
                 /usr/share/cmake-3.23/Modules/FindLAPACK.cmake
            >> apt-file search /usr/share/cmake-3.23/Modules/FindLAPACK.cmake
                 cmake-data: /usr/share/cmake-3.23/Modules/FindLAPACK.cmake
            >> ls /usr/share/cmake-3.23/Modules
                  ...  FindLAPACK.cmake ...
        ```

        You see here that `FindLAPACK.cmake` has been provided for you by kitware when installing `apt-get install cmake` wich trigger install of the `cmake-data` package: this is why `find_package(LAPACK)` works. There is _no_ `FindArpackNG.cmake` provided by cmake: this is why `find_package(ArpackNG)` will _not_ work.

Once again, I am not sure about all this, and, this is things like I understand them at this point: I may be wrong: my understanding is that case 1 stands to be "module mode" and case 2 stands to be "config mode" (https://cmake.org/cmake/help/latest/command/find_package.html#search-modes). My understanding is that a package can only provide "config mode" (this is what we do) but can't provide "module mode".

Would be good cmake to provide a `FindArpackNG.cmake` file for us: not sure how to achieve that !?... And then, I realise that we likely should mv `arpackng-config.cmake` into `ArpackNG-config.cmake`: the key given to `find_package` need to be the same (and in this case it would become "ArpackNG" instead of "arpackng" as it is now)

PR #376, opened 2022-10-21
bug fix: modify the condition for ierr=-3 in pxneupd.f

liuyangzhuan:
Change "else if (ncv .le. nev+1 .or.  ncv .gt. n) then" to "else if (ncv .le. nev+1 .or.  ncv .gt. nglo) then" in pxneupd.f before returning ierr=-3. nglo is the global number of matrix size, instead of local number of matrix size.

### DIFF
--- a/PARPACK/SRC/MPI/pcneupd.f
+++ b/PARPACK/SRC/MPI/pcneupd.f
@@ -316,7 +316,7 @@ c
      &           invsub, iuptri, iwev  , j    , ldh   , ldq   ,
      &           mode  , msglvl, ritz  , wr   , k     , irz   ,
      &           ibd   , outncv, iq    , np   , numcnv, jj    ,
-     &           ishift
+     &           ishift, nglo
       Complex
      &           rnorm, temp, vl(1)
       Real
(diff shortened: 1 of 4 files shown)


sylvestre:
Could you please add a test covering this case? thanks

liuyangzhuan:
@fghoussen I'm confused. You can see from
https://github.com/opencollab/arpack-ng/blob/8ecf1be71347f6b6eff5fc4f8c48e509cf3f4add/PARPACK/EXAMPLES/MPI/pdndrv1.f#L183 
https://github.com/opencollab/arpack-ng/blob/8ecf1be71347f6b6eff5fc4f8c48e509cf3f4add/PARPACK/EXAMPLES/MPI/pdndrv1.f#L305
that the local number nloc is used.

fghoussen:
>@fghoussen I'm confused.

Confused too! :) `arpack` is unfortunately not famous for it's clear documentation... It's known to be a black magic box...
Would say `n` must be equal or greater than the dimension of the problem. Your change seems to break everything...

liuyangzhuan:
@fghoussen Ok. You can decline this PR. I can just keep the changes locally as I do believe this fixes the problem I'm facing.

fghoussen:
@liuyangzhuan: feel free to close it or fix it! :)

PR #383, opened 2022-12-07
Better README

vickysharma0812:
## Pull request purpose

related to issue #382, and enhancement in documentation.

## Detailed changes proposed in this pull request

- README file in markdown form
- Please check the headings

PR #379, opened 2022-10-29
[CI] Windows job: run full test suite.

fghoussen:
## Pull request purpose

Checking if db55a7f enables to run full test suite on windows.

PR #384, opened 2022-12-10
Link to LAPACK before BLAS

FabienPean:
This PR swaps the order of linking to have LAPACK before BLAS.

Why ? When experimenting packaging the repo through vcpkg, I faced a peculiar problem when using static libraries (through WSL on Ubuntu 20.04 with GCC 9.4).

The first generated build command failed (with undefined references)

```bash
[1/2] /usr/bin/c++  -isystem /home/fabien/vcpkg/installed/x64-linux/include/arpack-ng -O2 -g -DNDEBUG -std=gnu++17 -MD -MT CMakeFiles/main.dir/main.cpp.o -MF CMakeFiles/main.dir/main.cpp.o.d -o CMakeFiles/main.dir/main.cpp.o -c ../../../main.cpp
[2/2] : && /usr/bin/c++ -O2 -g -DNDEBUG  CMakeFiles/main.dir/main.cpp.o -o main  /home/fabien/vcpkg/installed/x64-linux/lib/libarpack.a  /home/fabien/vcpkg/installed/x64-linux/lib/libopenblas.a  /home/fabien/vcpkg/installed/x64-linux/lib/liblapack.a  -lm  /usr/lib/x86_64-linux-gnu/libgfortran.so.5  -lgfortran  -lquadmath && :
FAILED: main 
```

By some chance it worked when fiddling around by manually adding targets in the CMakeLists of the test case

```bash
[1/2] /usr/bin/c++  -isystem /home/fabien/vcpkg/installed/x64-linux/include/arpack-ng -O2 -g -DNDEBUG -std=gnu++17 -MD -MT CMakeFiles/main.dir/main.cpp.o -MF CMakeFiles/main.dir/main.cpp.o.d -o CMakeFiles/main.dir/main.cpp.o -c ../../../main.cpp
[2/2] : && /usr/bin/c++ -O2 -g -DNDEBUG  CMakeFiles/main.dir/main.cpp.o -o main  /home/fabien/vcpkg/installed/x64-linux/lib/libarpack.a  /home/fabien/vcpkg/installed/x64-linux/lib/liblapack.a  -lm  /usr/lib/x86_64-linux-gnu/libgfortran.so.5  /home/fabien/vcpkg/installed/x64-linux/lib/libopenblas.a  -lgfortran  -lquadmath && :
```

The main difference was that `liblapack.a` was put before `libopenblas.a`. I tried the change given in this PR and it built successfully. 

It might not be the optimal solution, but the correct CMake alternative seems to be relatively recent (v3.24), see https://gitlab.kitware.com/cmake/cmake/-/issues/20078 and https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:LINK_LIBRARY

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,10 +386,10 @@ add_library(arpack ${arpackutil_STAT_SRCS} ${arpacksrc_STAT_SRCS} ${arpacksrc_IC
 
 target_link_libraries(arpack
   PUBLIC
-  $<INSTALL_INTERFACE:$<IF:$<BOOL:${BUILD_SHARED_LIBS}>,,BLAS::BLAS>>
   $<INSTALL_INTERFACE:$<IF:$<BOOL:${BUILD_SHARED_LIBS}>,,LAPACK::LAPACK>>
-  $<BUILD_INTERFACE:BLAS::BLAS>
+  $<INSTALL_INTERFACE:$<IF:$<BOOL:${BUILD_SHARED_LIBS}>,,BLAS::BLAS>>
   $<BUILD_INTERFACE:LAPACK::LAPACK>
+  $<BUILD_INTERFACE:BLAS::BLAS>
 )
 target_link_options(arpack PUBLIC "${EXTRA_LDFLAGS}")
 set_target_properties(arpack PROPERTIES OUTPUT_NAME arpack${LIBSUFFIX}${ITF64SUFFIX})


fghoussen:
@FabienPean: I get your point but this breaks CI so it seems something's wrong.
@juanjosegarciaripoll: can you help? Likely in connection with work previously done before merging #379.

FabienPean:
All the CI tests reporting failure are the MacOS ones and they failed on the step `Update OS`. This looks to me like something Github actions runners and/or MacOS specific, since it happened also for a documentation modification here: https://github.com/opencollab/arpack-ng/pull/383. 

It might be related to the transitioning `macOS-latest` to macOS 12 (https://github.com/actions/runner-images/issues/6384).

fghoussen:
> This seems random, so re-running the failed jobs might end up working. 

Didn't check logs! I'll re-run jobs, we'll see... If CI is broken, we will to fix this before merging anything else.

juanjosegarciaripoll:
The change looks correct to me. In my patches simply copied the order that was in the original CMakeLists.txt but this one does definitely makes more sense. In many systems, such as the ones I use, it makes no difference if openblas already includes lapack (which is often the case), which is why I did not notice a difference -- and probably why CI also did not.

FabienPean:
Here is the reason https://github.com/actions/runner-images/issues/2247
> Jobs using any of the macOS virtual environments may experience longer than normal queue times. In some cases it's possible workflows will time out, be cancelled, and need to be restarted.

I made a PR in my fork to be able to run the CI at will, it worked there for all without retry: https://github.com/FabienPean/arpack-ng/pull/1

fghoussen:
OK no hurry we can wait for github-actions to be back working!

fghoussen:
Good to merge if CI OK

PR #385, opened 2022-12-10
Make building tests an option in CMake

FabienPean:
Tests were built by default with CMake, it is now an option set to ON by default to preserve previous behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -622,11 +622,7 @@ endif()
 ############################
 # TEST
 ############################
-if(TESTS)
-  enable_testing()
-
-  set(CMAKE_CTEST_COMMAND ctest -V)
-
+function(build_tests)  
   set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/TESTS)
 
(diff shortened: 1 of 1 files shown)


fghoussen:
Can you put the previous code into a cmake function and call that function if the option is ON

FabienPean:
Is that the change you meant (should the function be defined there) ? It looks a bit weird, did you mean refactor the previous code to follow the "Examples" part of the CMakeLists instead ?

fghoussen:
Would be good to have equivalent of `function(examples list_name)` and `function(pexamples list_name)` but not sure if this can be easy done for all tests.

FabienPean:
```cpp
    enable_testing()
    set(CMAKE_CTEST_COMMAND ctest -V)   
```

had to be moved in global scope of the script because the CMake function creates a new scope. To keep it fully apart, it should be using the  `macro/endmacro` couple instead of the `function/endfunction`.

Now with regard to the CI, for the failing job `macos_latest_cmake_python`, it looks like a problem resulting from the transition to macOS 12.

fghoussen:
CI seems broken for other reasons: we'll have to fix it in a dedicated PR before merging this.

Issue #283, opened 2020-08-12
Document how to compile arpack-ng on Windows from C


eugene523:
I know it's not an issue, but can you present a more detailed documentation on how to compile it on Windows to use in c projects, especially for a person who is far away from Linux-related software and Fortran? By the way, visual studio project in VISUAL_STUDIO catalog is not opening in VS2015 on my computer.

fghoussen:
No idea. I'd say : 

- get LAPACK and BLAS here https://icl.cs.utk.edu/lapack-for-windows/lapack ? And then, add BLAS/LAPACK libs in the linker GUI of visual studio ?
-  use Unix ! :D MinGW or "Bash (Ubuntu) for Windows" and build / compile the unix way

If you get this done, feel free to PR how you did it in the doc (README.md)

fghoussen:
possible alternative : #303

dariomangoni:
I successfully built arpack on Win10 with VS2019 (16.10) and Intel MKL 2021.1.1 as **static** library.

To properly set all the environmental variables I highly recommend to use the _Intel oneApi command prompt for Intel x64 for Visual Studio_; from here call `cmake-gui` and follow the usual procedure.

Please mind to **uncheck** `BUILD_SHARED_LIBS`. From a quick look it seems that Fortran routines are missing the ATTRIBUTES DLLEXPORT that would be needed by exporting the function symbols into an import library. So you will end up building a .dll without any symbol exported and without its complementary import library (the .lib that usually comes with the .dll in Windows).

Then there is an additional issue on complex number declaration.
In Windows the `float _Complex` declaration does not work, so you have to manually change all this references:
- `float _Complex` to `_Fcomplex`
- `double _Complex` to `_Dcomplex`

I'm now able to run all the tests including `icb_arpack_cpp`, but not `icb_arpack_c` since there are some wrong (in my opnion) array initialization.

fghoussen:
Documentation has been provided here: https://github.com/opencollab/arpack-ng#windows-support

PR #391, opened 2023-01-19
Fix library consumption from C++17 on Windows

FabienPean:
Trying to use ARPACK on Windows from C++ with standard &geq; 17 fails because C `<complex.h>` content is replaced by C++ `<complex>` automatically unless the escape hatch macro is used `_CRT_USE_C_COMPLEX_H `

See https://github.com/MicrosoftDocs/cpp-docs/issues/4374 and https://github.com/microsoft/STL/issues/3280 for details.

### DIFF
--- a/arpackdef.h.in
+++ b/arpackdef.h.in
@@ -14,7 +14,9 @@
 #define a_uint unsigned int
 #endif
 #ifdef _MSC_VER
+#define _CRT_USE_C_COMPLEX_H
 #include <complex.h>
+#undef _CRT_USE_C_COMPLEX_H
 #define a_fcomplex _Fcomplex
 #define a_dcomplex _Dcomplex
 #ifndef CMPLXF


fghoussen:
@FabienPean : Thanks!
In separate PRs (not mixing up things !), would be appreciated if you could document how to build on windows (no idea how: I'm not windows-based - #283) or improve CI (we have only partial build/run test suite on windows - #379).

FabienPean:
I will try to have a look. Currently it is built on my machine through packaging with [vcpkg](https://github.com/microsoft/vcpkg) with a local test port. The goal is to pursue effort initiated there https://github.com/microsoft/vcpkg/pull/22803 and have a port merged. This will likely require a new patch version 3.8.1 if it goes well, and perhaps some other changes in the CMakeLists (about header files installed).

fghoussen:
@FabienPean: if you could document build on windows, put comments at a relevant place in the `README.md` for instance

PR #392, opened 2023-01-21
Run all examples for job `windows_latest_cmake` in CI

FabienPean:
Binaries of examples were not able to find `libarpack.dll` and failing, while tests were passing. The main difference in CMakeLists was setting the working directory for examples and not for tests. The binaries are probably linked without an absolute path for the library and searching the DLL in local folder. Changing the working directory in CMake changed the search local path where the DLL is absent.

To fix using MPI examples on Windows, the following was necessary:

* Add flag `-fallow-invalid-boz` fixing false positive errors on comments, e.g.
  * >    87 | c  1. D.C. Sorensen, "Implicit Application of Polynomial Filters in
                Warning: missing terminating " character 
* Replace `mpirun` by standard `mpiexec`
* Fix working directory of tests for MPI binaries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,7 +212,10 @@ if (MPI)
         endif()
     endif()
 
-    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${MPI_Fortran_COMPILE_FLAG} -fallow-invalid-boz")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${MPI_Fortran_COMPILE_FLAG}")
+    if(CMAKE_SYSTEM_NAME MATCHES "Windows" AND CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+        set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fallow-invalid-boz")
+    endif()
 
     # Check if we can use ISO_C_BINDING provided by MPI.
     file(WRITE ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/PROG_ICB.f90
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ function(pexamples list_name)
         get_filename_component(lwe ${l} NAME_WE)
         add_executable(${lwe} ${parpackexample_DIR}/${l} )
         target_link_libraries(${lwe} parpack arpack MPI::MPI_Fortran)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ function(examples list_name)
         get_filename_component(lwe ${l} NAME_WE)
         add_executable(${lwe} ${arpackexample_DIR}/${l} ${examples_EXTRA_SRCS})
         target_link_libraries(${lwe} arpack BLAS::BLAS LAPACK::LAPACK ${EXTRA_LDFLAGS})
-        add_test(NAME "${lwe}_ex" COMMAND ${lwe} WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+        add_test(NAME "${lwe}_ex" COMMAND ${lwe})
```
fghoussen, line 66: Really?!... That was the only thing to modify to get full build/tests run OK on windows?! 
Can you elaborate/explain a bit why?

fghoussen:
Thanks! Can you update the `CHANGES` log? And also add maybe a note there related to #391 (forgot to ask to update change log before merging #391)

fghoussen:
@FabienPean: just tried to push on top of you to see if we could also get MPI build/test...

CMakeLists.txt
FabienPean: The search for the DLL is not an absolute path somehow, so it was searching locally for the library. Setting a different working directory meant it was searching for the library there. Since it was absent, it then returned the error code 0xc0000135 indicating DLL not found.

FabienPean:
The CI now also runs the MPI examples and the changelog has been updated.

CMakeLists.txt
```diff
@@ -212,7 +212,7 @@ if (MPI)
         endif()
     endif()
 
-    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${MPI_Fortran_COMPILE_FLAG}")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${MPI_Fortran_COMPILE_FLAG} -fallow-invalid-boz")
```
fghoussen, line 215: Is it safe to add `-fallow-invalid-boz` by default?
Should it be added under some kind of test like "if COMPILER = GNU and OS = WINDOWS"?
Opened question! Not sure what is best

FabienPean: That's a good point, I just followed the compiler complaint without much second thought 😄
Though, what bothers me is why the compiler generated errors for lines which are comments in the first place... I am not familiar with Fortran so hard to say

fghoussen: > That's a good point,

OK, add this to the PR, and it's good to go for me.

FabienPean: Done. It seems this is related to the code being old Fortran 77

fghoussen:
@FabienPean: merci Fabien ! :)

FabienPean:
Avec plaisir :)

PR #393, opened 2023-01-26
Avoid removing macro `_CRT_USE_C_COMPLEX_H` if pre-existing

FabienPean:
Last changes unilaterally undefined the macro no matter if it existed before. To keep it transparent and without side effects for anyone potentially relying on it in some other ways, it is now undefining the macro only if it was not previously defined.


fghoussen:
@FabienPean: thanks!
Would be great if you could add a few lines in `README.md` to explain how to compile on windows (using `vcpkg` I guess?).

FabienPean:
Yes I can do that but only after I get the port to be merged upstream. One thing at a time :)

Issue #401, opened 2023-02-16
DSAUPD() fails with ARPACK 3.9.0, but not with ARPACK 3.8.0


szhorvat:
This is a summary of the issue described at https://github.com/igraph/igraph/issues/2311, i.e. DSAUPD() failing when running the igraph test suite with ARPACK 3.9.0. With ARPACK 3.8.0, there is no problem.

**We are solving the following eigenvalue problem:**

 - Find the leading eigenvalue/vector of an $n\times n$ matrix $A$ where $A_{ij} = 0$ for all $i,j$ except that $A_{12} = A_{21} = 1$. The problem occurs specifically with $n=10$ and $n=20$, not other values.
 - Find a single eigenvalue/vector using `which='LA'`

**The error**

An error occurs depending on the starting vector that is used. The error is easy to reproduce after trying a few random starting vectors.

Details of the error:

 - Error flag `-9999` is returned in `info`.  The value of`iparam(5)` is `4` when $n=10$. See https://github.com/opencollab/arpack-ng/blob/master/SRC/dsaupd.f#L273

**Reproducing**

The issue can be reproduced independently of igraph, using Octave. This suggests that the problem may be in ARPACK itself, and not with how igraph uses ARPACK.

Transcript of Octave session:

```matlab
octave:20> n=10 # problem occurs with n=10 or n=20 only
octave:21> A=zeros(n,n); A(1,2)=A(2,1)=1
A =

   0   1   0   0   0   0   0   0   0   0
   1   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0

octave:22> for i=1:100; opt.v0 = rand(n,1); [v,d]=eigs(A,1,'la',opt); end
error: eigs: error in dsaupd: 
error: called from
    eigs at line 338 column 22
```

The purpose of the `for` loop is to find a random starting vector that reproduces the issue.

[...]

Issue #215, opened 2019-06-29
For zero input matrix, fails with "starting vector is zero"


pv:
## Expected behavior

With ARPACK < 3.6.0 this works:
```
$ octave
octave:1> m = zeros(4,4);
octave:2> eigs(m, 1)
ans = 0
```

## Actual behavior

With ARPACK 3.7.0:
```
$ octave
octave:1> m = zeros(4,4);
octave:2> eigs(m,1)
error: __eigs__: eigs: error in dsaupd: Starting vector is zero
This should not happen.  Please, see https://www.gnu.org/software/octave/bugs.html, and file a bug report
error: called from
    eigs at line 285 column 20
```

## Notes, remarks

This is probably from the same changes as #142: forcing the initial vector to the range of a zero matrix produces a zero initial vector, which fails. The `*getv0` routines could perhaps check if `resid` is a zero vector, and (in the non-generalized cases?), skip the forcing to the range?

caliarim:
Yes, the problem could be related with #79 (force the initial vector to be in the range). In #142 there is a further modification which should manage rank deficient matrices. This case is the most rank deficient matrix. I think it should be enough simply not to check whether the initial vector is zero (around line 324 of dnaup2.f and related). In fact, after the first try, the initial vector is no more forced to be in the range (this was the outcome of #142).

By the way, this issue should appear also with 3.5.0, by selecting the generalized problem with the zero input matrix and the identity matrix B. But you need plain arpack to check, not octave.

caliarim:
I verified that arpack 3.5.0 with a dummy generalized problem with the zero input matrix and the identity matrix B stops with "starting vector is zero".

QuLogic:
Any ideas here? This is also breaking some tests when [building the R igraph package](https://koji.fedoraproject.org/koji/taskinfo?taskID=41797324) with arpack 3.7.0.
```
  -- 1. Error: cluster_leading_eigen works (@test_leading.eigenvector.community.R#
  At arpack.c:944 : ARPACK error, Starting vector is zero
  Backtrace:
   1. igraph::cluster_leading_eigen(g)
  ```

**Comment by caliarim** [comment]
I think I found a workaround at the time, give me a week or so.

**Comment by caliarim** [comment]
This is my proposal
```
diff --git a/SRC/dgetv0.f b/SRC/dgetv0.f
index 1d6dc01..9635720 100644
--- a/SRC/dgetv0.f
+++ b/SRC/dgetv0.f
[...]

PR #408, opened 2023-02-25
[BUG FIX] arpackmm: fix restart.

fghoussen:
* Fixing issue401.

* Fixing issue215 with shift.


szhorvat:
It seems to me that this PR only makes changes to the arpackmm interface, but not to ARPACK itself. Issue #401 is unrelated to arpackmm. Am I missing anything?

szhorvat:
I see that you added a test with the matrix from #401, but I don't see how the starting `resid` is set. #401 is reproducible with certain (reasonable) starting `resid` vectors. Can you try these? See https://github.com/opencollab/arpack-ng/issues/401#issuecomment-1435530450

fghoussen:
> It seems to me that this PR only makes changes to the arpackmm interface, but not to ARPACK itself.

Correct. To me the 401/215 issues are not related to arpack, but in the misuse of it (in arpackmm, igraph, octave) like it often happens as the doc is no-great!... Indeed, this PR fixes arpackmm only

> Issue #401 is unrelated to arpackmm. Am I missing anything?

Correct. #215 is unrelated to #401, but, on the way to fixing #401, I had the same issue than #215, so I had to look at that too. In the end, I was able to get arpack to work on both #401 and #215 following the guidelines explained here: https://github.com/fghoussen/arpack-ng/tree/arpackmm#faq.

My understanding (may be wrong !...) is that:

- in #401, it looks like you set your initial vector `v` in the wrong place https://github.com/opencollab/arpack-ng/issues/401#issuecomment-1439075297: `v` should not be placed in `resid`. Starting from here, various BLAS on vaious distros may end up in different behavior depending on numerical values you have at the initialization. 

- in #215, I guess `resid ` was set to zero so that `info = -9 Starting vector is zero` is to be expected and is actually 100% logical! :) As `A` is zero and the eigen value to find is zero too, you *need* to shift just a bit in order to start arpack with something which is *not* zero (i.e. with a `resid = Av - lamdba*v` which is not zero at start). This looks to me like a very specific (corner ?! :) ) case. I guess here too depending on numerical  values you have at the initialization, you may succeed or not to start arpack.


> I see that you added a test with the matrix from #401, but I don't see how the starting `resid` is set. 

For from-scratch runs:
https://github.com/fghoussen/arpack-ng/blob/99bc00e1b18cfce3eb0248b4375852b5a4174776/EXAMPLES/MATRIX_MARKET/arpackSolver.hpp#L734

For restart runs:
https://github.com/fghoussen/arpack-ng/blob/99bc00e1b18cfce3eb0248b4375852b5a4174776/EXAMPLES/MATRIX_MARKET/arpackSolver.hpp#L684

> #401 is reproducible with certain (reasonable) starting `resid` vectors. Can you try these? See [#401 (comment)](https://github.com/opencollab/arpack-ng/issues/401#issuecomment-1435530450)

You can do *all* tests (likely more than one!) you'll need by yourself:

- run arpackmm once: this will produce `arpackSolver.resid.out` and `arpackSolver.v.out` as explained in `arpackmm --help`
- Put whatever values you want to test in `resid` and `v` through theses files and re-run arpackmm with restart option `arpackmm --restart ..... --verbose 3` and maybe verbose option: restart with your initial vector.

All tests I did at the command line seems to work now as I managed `arpackmm` to never let `resid` to zero (lesson learned from #215)

fghoussen:
> in https://github.com/opencollab/arpack-ng/issues/215, I guess resid was set to zero so that info = -9 Starting vector is zero is to be expected and is actually 100% logical! :) As A is zero and the eigen value to find is zero too, you need to shift just a bit in order to start arpack with something which is not zero (i.e. with a resid = Av - lamdba*v which is not zero at start).

Another possible reason for 215 issue may be `v` init that in this cas must *not* be exactly zero but some small values: https://github.com/fghoussen/arpack-ng/blob/99bc00e1b18cfce3eb0248b4375852b5a4174776/EXAMPLES/MATRIX_MARKET/arpackSolver.hpp#L741 

The problem is that they are lots of misleading examples out there on how you could use arpack... Not always clear to understand what to do. `for (a_int n = 0; n < ldv*nbCV; n++) v[n] = zero; // Avoid "bad" starting vector.` is something I got from one of these samples... In #215 setting *both* `resid` and `v` to *exactly* zero means `resid` is zero at start and will prevent arpack from starting...

fghoussen:
Seems `make distcheck` fails: OK here, no idea why!?... Any clue is welcomed

Issue #411, opened 2023-03-13
Different RNG seeding strategies in complex vs real versions of parallel getv0()


szhorvat:
I noticed that in PARPACK, the initial seed is chosen differently in Z/C-prefix complex functions vs D/S-prefix real ones.

In `pcgetv0` (and `pzgetv0`), we have:

https://github.com/opencollab/arpack-ng/blob/master/PARPACK/SRC/MPI/pcgetv0.f#L218

```fortran
c     %-----------------------------------%
c     | Generate a seed on each processor |
c     | using process id (myid).          |
c     | Note: the seed must be between 1  |
c     | and 4095.  iseed(4) must be odd.  |
c     %-----------------------------------%
c
      call MPI_COMM_RANK(comm, myid, ierr)
      igen = 1000 + 2*myid + 1
      if (igen .gt. 4095) then
         write(0,*) 'Error in p_getv0: seed exceeds 4095!'
      end if
c
      iseed(1) = igen/1000
      igen     = mod(igen,1000)
      iseed(2) = igen/100
      igen     = mod(igen,100)
      iseed(3) = igen/10
      iseed(4) = 7
```

I.e., each parallel thread gets a unique seed, making sure they don't generate the same sequence of random numbers.

----

But in `psgetv0` / `pdgetv0`, we have the same fixed initial seed as in the serial version:

https://github.com/opencollab/arpack-ng/blob/master/PARPACK/SRC/MPI/pdgetv0.f#L215

```fortran
c     %-----------------------------------%
c     | Initialize the seed of the LAPACK |
c     | random number generator           |
c     %-----------------------------------%
c
      iseed(1) = 1
      iseed(2) = 3
      iseed(3) = 5
      iseed(4) = 7
```

This does not look correct to me. The seed should be different for each parallel thread, just like in the `Z`/`C` versions.

----

This is not related to the earlier seeding issues I reported. I do not use PARPACK myself, this is just something I noticed while looking at the code.

This goes back to the original PARPACK, and is not something that was changed in ARPACK-NG. To be accurate, the setting of `iseed(4)` _was_ changed (unnecessarily, as it was already odd), but the general setup wasn't.

Note that the original PARPACK came as two archives: the original code, and a patch. Only the patch has the special seeding based on the process ID. It seems to me that they simply forgot to include the `D`/`S` versions in the patch.

szhorvat:
Also, this does not look right:

```fortran
      if (igen .gt. 4095) then
[...]

Issue #410, opened 2023-03-11
Initial vector generation is broken in 3.9.0, causing rare failures


szhorvat:
In ARPACK, [the `Xgetv0()` function](https://github.com/opencollab/arpack-ng/blob/master/SRC/dgetv0.f) function generates a random initial vector for a numerical iteration process. With low probability, it may occur that this random vector won't be suitable for the numerical algorithm. In this case, ARPACK re-generates it, up to three times, until a suitable one is found. This happens here:

https://github.com/opencollab/arpack-ng/blob/master/SRC/dsaitr.f#L409

Commit ce2e69a849da1d10dad5d6d3ec4db6120b3ecf50 breaks this process by making `Xgetv0()` deterministic. The only change the commit makes is to re-seed the random number generator with the very same fixed seed at the beginning of `Xgetv0()`. Therefore, if with this fixed seed `Xgetv0()` happens to produce an unsuitable "random" vector, all subsequent retries will obviously fail in the same way.

The fact that the above-linked code performs random re-tries until success makes it obvious that ce2e69a849da1d10dad5d6d3ec4db6120b3ecf50 (which removes the randomness) is incorrect.

----

I explained this already in detail in #401. This issue is to make the problem plain to see without having to go through the long discussion there. The description is kept intentionally simple.

----

While the failure occurs with low probability, and it is dependent on the LAPACK implementation, it does occur in practice. It is currently causing the igraph test suite to fail with ARPACK 3.9.0 on several platforms ([see e.g. with msys2](https://github.com/igraph/igraph/actions/runs/4388756464/jobs/7685610006)). I confirmed that the failure mode is precisely the one described above, by adding `print` statements to follow the code path, and verify that inside `Xgetv0()` the very same "random" vector keeps getting generated again and again. A direct revert of ce2e69a849da1d10dad5d6d3ec4db6120b3ecf50 eliminates the problem.

fghoussen:
Would be good, if possible, to have reproducible / deterministic data to reproduce the problem.

szhorvat:
Can you address the obvious logic flaw that I described above?

szhorvat:
This issue should be obvious from following the logic of the code. It would be good if you could clarify which part of the description needs clarification, and what sort of reproducer will provide this.

[...]

PR #414, opened 2023-04-09
fix: ensure that LAPACK RNG state is propagated

### LINKED ISSUE #401: DSAUPD() fails with ARPACK 3.9.0, but not with ARPACK 3.8.0

szhorvat:
- Fixes #401, #410, #411.
 - Restores 'inits' variable removed in ce2e69a849da1d10dad5d6d3ec4db6120b3ecf50, ensuring that the RNG state is propagated
 - Reverts e0d67054f573da351f12a226f7c7cc65a690ef3d to ensure that seed is different on each parallel thread. Note that the seed is set based `igen` (computed from the process ID), the least significant digits of which was discarded by the mentioned commit, potentially making the seed identical for several parallel workers. This commit tried to ensure that `iseed(4)` was odd, but this was already ensured as `igen` was computed to be odd.
 - Updates seed initialization of parallel pdgetv0/psgetv0 so that they match that of pzgetv0/pcgetv0. This last change is made to code inherited from the original ARPACK, not new code written for ARPACK-NG. Please see #411 for the explanation.


szhorvat:
Note: I won't be able to respond to comments until April 19th Wednesday.

fghoussen:
Thanks!

Can you add comments/remarks in each file regarding lessons learned? (why-when-how the init is supposed to work? why several v0 init and if which cases, etc - arpack is maintained only, the doc is ugly, initial developpers are no more maintaining this... So all information are precious here [personal point of view])

Also, can you update the change log? 

> I won't be able to respond to comments until April 19th Wednesday.

That's OK. No hurry!

sylvestre:
could you please add tests? thanks

szhorvat:
Let's please avoid force pushes until the PR is ready for merging, just to make my life easier. I can squash or rearrange all commits before merging.

fghoussen:
Take your time, no hurry! :)

> Just to confirm, the test failures I see here are not expected, right?

Nope!

> Maybe it's better to try to separate the fixes for https://github.com/opencollab/arpack-ng/issues/410 and https://github.com/opencollab/arpack-ng/issues/411.

Yes!

> Let's please avoid force pushes until the PR is ready for merging, just to make my life easier.

Sure! I updated the branch with github GUI to see if failures were real (jobs may stuck on OS upgrade).

fghoussen:
If you intent to separate the fixes, do you need to keep this PR opened?

szhorvat:
Please let's keep this open. It'll only include the first part. Unfortunately I probably won't have time for the second part this summer. I'll clean up the commits before they're ready for merging.

fghoussen:
OK, no problem! :)

szhorvat:
This PR now only has the fix for #410.

Can you please approve the CI run?

I am hoping that this can be merged, since it continually affects igraph. I am severely overwhelmed at the moment, and I won't have time to add tests for it. I will note that on my machine the test suite fails on `master`, without this change:

```
	 73 - bug_1315_single_tst (Failed)
	 81 - icb_arpack_cpp_tst (Subprocess aborted)
```
 Built with these options:

```
EIGEN                            OFF                                                                        
 EXAMPLES                         ON                                                                         
 ICB                              ON                                                                         
 INTERFACE64                      OFF                                                                        
 ITF64SUFFIX                                                                                                 
 LIBSUFFIX                                                                                                   
 MPI                              OFF                                                                        
 PYINT                            int32                                                                      
 PYTHON3                          OFF                                                                        
 SYMBOLSUFFIX                                                                                                
 TESTS                            ON
```

fghoussen:
> Can you please approve the CI run?

CI passed. Is the PR complete?

szhorvat:
> CI passed. Is the PR complete?

Yes, for the serial part of ARRPACK (i.e. excluding PARPACK) the PR is complete. I suggest looking at PARPACK separately.

> As the patch seems light, do you think you'll have time to impact PARPACK too (in this PR or another) to keep things symmetric?

Sure, we can do that. The previous version of this PR was doing mostly that, plus #411, which is making the different versions of (complex vs real) symmetric. I'll push that as a separate commit, so we can look how it affects tests.





> Also, if PR is complete, can yo update the change log?

No problem.

----

I'll be travelling for a week, I'll see if I can do these while on the train. Hoping for good WiFi.

szhorvat:
All this is done:

 - Updated seeding strategy in PARPACK (for symmetry)
 - Made sure that seeding strategy is symmetrical across Z/C and D/S versions in PARPACK (fixes #411)
 - Updated changelog

I cannot test PARPACK while travelling, as I don't have access to MPI. Would you be able to check that the 2nd and 3rd commits don't cause any new test failures, @fghoussen ?

fghoussen:
> I cannot test PARPACK while travelling, as I don't have access to MPI.

CI will test us!

fghoussen:
[7cec434](https://github.com/opencollab/arpack-ng/pull/414/commits/7cec434acbd0114730b9f04c6245cec9ac5d7392) fails but [e60fe19](https://github.com/opencollab/arpack-ng/pull/414/commits/e60fe191a8544a29085b2b5732027c5bfe936850) pass... and the diff is changelog?!...

I'll have a look when possible.

szhorvat:
The parallel version is not deterministic because the initial seed depends on the process ID of parallel threads. This is done to make sure that the different threads generate a different sequence of random numbers. This is of course annoying in a testing context, but we don't have control over the process IDs ...

In contrast, the serial version uses the same seed in every run so it is deterministic.

fghoussen:
> This is of course annoying in a testing context

A way to fix this must be found: stable CI is mandatory!

szhorvat:
I agree that it'd be great to improve on this, but I do not have the Fortran and MPI expertise, and I truly don't have the time to learn what's needed (in the middle of an overseas move and taking on new commitments). The seeding strategy that's used here is not from me, neither from ARPACK-NG, but from the original ARPACK. Or rather: from the extra patch that's available in addition to the original ARPACK archive.

It uses `MPI_COMM_RANK`, and since I have no MPI experience, I am not even sure what this is. A bit of research suggests that this is not quite the PID, but it's numbered from 0 to `MPI_COMM_SIZE - 1`.

However, in a parallel random setting, there is not much one can do to ensure determinism. Deterministic seeding is insufficient, as the time the different threads/processes take to complete their chunk of work is unpredictable. _In general_, the next chunk of work will go to the thread/process that finishes first.

This has been an issue for us in igraph as well, which includes a parallelized random sampling algorithm. The results are not deterministic, even if the number of threads is fixed, and the initial seeds are set consistently.  That's because the order in which the different threads take on work depends on factors that we cannot control. Since makes testing at least inconvenient, thus we resorted to testing this algorithm with a single parallel thread.

szhorvat:
Would you perhaps be willing to merge the fix only for the serial version, and leaving the parallel one for later? I'm happy to separate this into two PRs.

Once again, I do believe the parallel fix to be correct, but with no MPI expertise and little time, I won't be able to work on the parallel version in the near term. I am hoping that this won't block the serial fix.

fghoussen:
> The parallel version is not deterministic because the initial seed depends on the process ID of parallel threads.
>  in a parallel random setting, there is not much one can do to ensure determinism. 

I understand your point, but, we need robust CI or it's useless!

> Deterministic seeding is insufficient, as the time the different threads/processes take to complete their chunk of work is unpredictable.

AFAIU, each MPI proc handles a independent part of work https://github.com/opencollab/arpack-ng/pull/414#issuecomment-1501563595?
Not sure here, but, if it is, should deterministic seeding be enough to make CI robust? 

> It uses MPI_COMM_RANK, and since I have no MPI experience, I am not even sure what this is. A bit of research suggests that this is not quite the PID, but it's numbered from 0 to MPI_COMM_SIZE - 1.

You guessed right!

> That's because the order in which the different threads take on work depends on factors that we cannot control.

Correct. That's a tricky problem if work assigned to threads are dependent.

PARPACK/SRC/MPI/pcgetv0.f
```diff
@@ -213,26 +219,30 @@ subroutine pcgetv0
...
+         iseed(3) = igen/10
+         iseed(4) = mod(igen,10)
 c
-      iseed(1) = igen/1000
-      igen     = mod(igen,1000)
-      iseed(2) = igen/100
-      igen     = mod(igen,100)
-      iseed(3) = igen/10
-      iseed(4) = 7
+         inits = .false.
```
fghoussen, line 244: AFAIU, once `inits` is back to `.false.`, nothing makes it back to `.true.`!?...
"Who" can or need to make `inits` back to `.true.`, when and why?!... Not clear to me: could you detail this https://github.com/opencollab/arpack-ng/pull/414#issuecomment-1501563595 (and possibly push comments about this in the code to keep track of the logic)? Once this is clear to me, I may have a look at the MPI part and find a way to make CI robust.

szhorvat: The purpose of `inits` is to ensure that the RNG state is set only once, and after that it is propagated correct. `inits` will be `.true.` on startup. Then as soon as the RNG is initialized/seeded, it is set to `.false.` to prevent future re-initialization. It will never be set to `.true.` again during the program's life time.

This logic is from the original ARPACK.

Can we please merge the fix for the serial part before handling the parallel one? I can separate the parallel one into a different PR. Merging the serial fix is a priority for us.

fghoussen: I'd like to find a way to fix this for both serial/parallel otherwise parallel will never be fixed (not done today = never done...). I'll have a look when possible.

fghoussen:
Fixed by #423

PR #416, opened 2023-04-21
Fix windows icb macros

Vonfry:
## Pull request purpose

fixing issues in icb header.

- I macro is defined by complex.h on windows which is also defined in arpackSolver. IMO, this macro is useless, so it is undef.
- Distinct icx from cl on windows because icx uses clang-cl which defined _MSC_VER as well.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 arpack-ng - 3.9.1
 
+[ Ruoyu Feng ]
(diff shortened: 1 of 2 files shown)


fghoussen:
Can you update the change log?

Vonfry:
@fghoussen CHANGES was updated. Would you like to check it?

PR #419, opened 2023-05-26
Fix package name in CMake find_package() examples

aradi:
## Pull request purpose

Fixing package name typos in the README, where the package search via CMakes `find_package()` is demonstrated.

Use `arpackng` as package name instead of `arpack-ng`, as CMake won't find the installed Arpack-library with the latter.


fghoussen:
Indeed...

https://github.com/opencollab/arpack-ng/blob/3e3e12fef2c516fe21658d2e67dc8cbe86ec60bb/cmake/tstCMakeInstall.sh.in#L54

Thanks!

PR #423, opened 2023-08-02
Fix `icb_parpack_c` and `icb_parpack_cpp` to split the problem size across MPI processes

FabienPean:
I had a bit of spare time and felt like doing a bit of archeology, so I went digging into parpack issues raised in #414 

The story in short is that all tests were only succeeding with an identical seed array was set for all processes. Noticeably, the error occurred only in icb_parpack_c/cpp and not in Fortran parpack tests indicating a discrepancy in the test.

The issue is that the problem was not split across processes in the test file before calling to PARPACK. This effectively made all processes computing the whole problem and incurring weird behavior due to each process having a different initial vector but trying to share data through MPI. 

Changes are contained in the last commit https://github.com/opencollab/arpack-ng/commit/d53f32aa84dc3c03c5d50615c10fd6b466d367ec built atop #414


fghoussen:
@FabienPean: thanks! Indeed I think you found the root cause of all this.
I'll have a closer look by the week-end, but, it sounds right!

fghoussen:
@FabienPean: could you add a bullet point in the FAQ of the README.md that would say something like
```
  - Using PARPACK, I get incorrect eigen values.

    Make sure each MPI processor handles a subpart of the eigen system (matrices) only.
    ARPACK handles and solves the whole eigen problem (matrices) at once.
    PARPACK doesn't: each MPI processor must handle and solve a subpart of the eigen system (matrices) only (independently from the other processors).

```

FabienPean:
The comment to the FAQ has been added, with an extra hint on where to find examples how to do it.

fghoussen:
@szhorvat, @FabienPean: thanks!

PR #424, opened 2023-08-02
Set a true random seed for LAPACK RNG

### LINKED ISSUE #411: Different RNG seeding strategies in complex vs real versions of parallel getv0()

FabienPean:
Depends on #423
Fix #411 

With the test fixed, this opens the possibility to use a fully random seed for each MPI processes (and also to the serial version). 

I don't know the arcane of Fortran77, so I searched and decided to use the (non-standard?) [`LOC`](https://gcc.gnu.org/onlinedocs/gfortran/LOC.html) (also mentioned [here](https://docs.oracle.com/cd/E19957-01/805-4939/z40007368469/index.html), [here](https://www.ibm.com/docs/en/xl-fortran-aix/16.1.0?topic=procedures-locx-extension), and [here](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/loc-002.html)) intrinsic to get addresses of local variables as a source state for the seed. 

Is there a better alternative? Perhaps one based on time but [`TIME`](https://gcc.gnu.org/onlinedocs/gcc-4.2.4/gfortran/TIME.html) does not seem standard in Fortran77 too. Otherwise, the seed could be provided by the user but it means bubbling up the parameter and affecting function signatures. Or simply keeping it fixed like before.

Anyway, modifications are contained in commit https://github.com/opencollab/arpack-ng/pull/424/commits/f7d35114e6b0d456f434b467d8a64dff0e7bd688


FabienPean:
The change affects some jobs: `arpackmm.sh` for ubuntu_latest_autotools, and `pyarpackSparseCGILU_tst` for macos_latest_cmake_python. I hadn't checked them, putting back to draft for now.

fghoussen:
> use a fully random seed 

Not sure this is worth the effort unless you really want this and can afford more debug on your spare time... ;)
I mean arpack is an old thing that it is often simpler to touch with minimal impact and only when needed... Unless you fall into another rabbit hole...

The only thing I would keep are all impacts of the family:
```
>> git diff
diff --git a/SRC/dgetv0.f b/SRC/dgetv0.f
index 8be4fa2..aaf8225 100644
--- a/SRC/dgetv0.f
+++ b/SRC/dgetv0.f
@@ -196,7 +196,7 @@ c
       iseed(1) = 1
       iseed(2) = 3
       iseed(3) = 5
-      iseed(4) = 7
+      iseed(4) = 7*2+1
 c
       if (ido .eq.  0) then
 c
``` 
As it's explicitly stated that `ISEED(4) must be odd` in https://netlib.org/lapack/explore-html/df/dd1/group___o_t_h_e_rauxiliary_ga77e05a87ced667cbdb502aa87c72d056.html: I guess if you can push this, we're all good!

fghoussen:
> > use a fully random seed
> 
> Not sure this is worth the effort unless you really want this and can afford more debug on your spare time... ;) I mean arpack is an old thing that it is often simpler to touch with minimal impact and only when needed... Unless you fall into another rabbit hole...
> 
> The only thing I would keep are all impacts of the family:
> 
> ```
> >> git diff
> diff --git a/SRC/dgetv0.f b/SRC/dgetv0.f
> index 8be4fa2..aaf8225 100644
> --- a/SRC/dgetv0.f
> +++ b/SRC/dgetv0.f
> @@ -196,7 +196,7 @@ c
>        iseed(1) = 1
>        iseed(2) = 3
>        iseed(3) = 5
> -      iseed(4) = 7
> +      iseed(4) = 7*2+1
>  c
>        if (ido .eq.  0) then
>  c
> ```
> 
> As it's explicitly stated that `ISEED(4) must be odd` in https://netlib.org/lapack/explore-html/df/dd1/group___o_t_h_e_rauxiliary_ga77e05a87ced667cbdb502aa87c72d056.html: I guess if you can push this, we're all good!

According to this remark https://github.com/opencollab/arpack-ng/issues/411#issuecomment-1466191391, in MPI fortran files `iseed(4)` should already be odd, so, this change should not be necessary but could make things more explicit IMO.

fghoussen:
@FabienPean: do you wanna go to the end of this?

FabienPean:
Haven't had more time for it yet. I will try to see if there is a common pattern in the failures, as it seems suspiciously focused on a few tests. Otherwise, I will revert it to a fixed seed and add a comment in the FAQ to advice users to give themselves an initial guess (if it is not there). Let's keep it as a draft at the moment.

fghoussen:
@FabienPean: OK, I let you decide if you want to push this to the end or not! ;)

FabienPean:
After running the tests on a sizeable amount of repetitions. With a random seed, a few tests fail (by order of occurrence):
1. pyarpackSparseCGILU_tst (>20%)
2. pyarpackRestart_tst (~5%)
3. bug_1315_single_tst (~1%)

Further details:

1) For pyarpackSparseCGILU_tst, the test is failing [on the second part](https://github.com/opencollab/arpack-ng/tree/master/EXAMPLES/PYARPACK/pyarpackSparseCGILU.py.in#L116-L117). The CG solver is called on a non-hermitian matrix, which is not guaranteed to succeed. Testing it with any BiCG will (always) pass.

2) For pyarpackRestart_tst, the test will fail with a bad seed [here](https://github.com/opencollab/arpack-ng/blob/c18d153f20ac9a4c5e6e4aa9023c4f7fa5c9ada6/EXAMPLES/PYARPACK/pyarpackRestart.py.in#L107), only because the restart requires more iterations than in the initial case, otherwise always converge. (If it is seed dependent, does it make sense to strictly require better number of iterations on a single input?)

3) For bug_1315_single_tst, it is a [tolerance check not passing](https://github.com/opencollab/arpack-ng/blob/c18d153f20ac9a4c5e6e4aa9023c4f7fa5c9ada6/TESTS/bug_1315_single.c#L75). Wouldn't it be better to use a relative error instead of an absolute one on the checks?

4) Furthermore, the test failing on `ubuntu_latest_autotools` is the call to `./arpackmm.sh`, yet the cmake/ctest counterpart on other job `ubuntu_latest_cmake` passes. For this I have no clue what would be the source of discrepancy.

Anyway some of those point are not directly solvable, so I will close the issue, as the pre-selected seed value seems the best compromise. Though comments on those points above would be appreciated @fghoussen

fghoussen:
>    pyarpackSparseCGILU_tst (>20%)
>    pyarpackRestart_tst (~5%)
>    bug_1315_single_tst (~1%)

Interesting! If possible, do not hesitate to propose a PR for each of them

> Wouldn't it be better to use a relative error

As this is float, maybe we could test absolute error, then relative error and break if both fails for s/c flavors?...
Others examples `icb_*, *.c[pp]` are built on the same pattern: would be good to make this change everywhere... 

> Though comments on those points above would be appreciated

They are! :)

PR #398, opened 2023-02-11
[BUG FIX] Fix install: headers in /path/to/local/include/arpack

fghoussen:
## Pull request purpose

fixing issue #394

PR #426, opened 2023-08-13
Fix header and pkg-config file installation

turboencabulator:
This reverts some undesirable changes introduced in the 3.9.0 release:

- Finish changing the include directory from `${includedir}/arpack-pc` back to `${includedir}/arpack` (started in #398 but was missing some files).
- Don't install parpack.pc and arpackSolver.pc (or their suffixed variants) unless those libraries were actually built.


turboencabulator:
@fghoussen Done.  Looks like you've already fixed a couple of the same issues I found.

I did just notice that CMake is installing arpackicb.h even when the ICB option is not enabled, is that a bug?

fghoussen:
I'll have a closer look when possible.
Can you move back pc files in pkg-config dir? This way all *.cmake/*.pc files (dealing with finding cmake-package or pkg-config-module) are at the same level in a dedicated directory with associated tests (sh)

> @fghoussen Done. Looks like you've already fixed a couple of the same issues I found.

Yep ! :)

> I did just notice that CMake is installing arpackicb.h even when the ICB option is not enabled, is that a bug?

AFAIR, `arpackicb.h`is meant to define types for fortran subroutines

fghoussen:
> I did just notice that CMake is installing arpackicb.h even when the ICB option is not enabled, is that a bug?

I think arpackicb.h can be installed only when OCB=ON. [afd8413](https://github.com/opencollab/arpack-ng/pull/426/commits/afd8413934b32c3f19c697211cde2af4114954ab) will tell...

turboencabulator:
> Can you move back pc files in pkg-config dir? This way all _.cmake/_.pc files (dealing with finding cmake-package or pkg-config-module) are at the same level in a dedicated directory with associated tests (sh)

I looked into a couple different ways to do it and I wasn't happy with either.  The makefiles would reference things outside of their directory, or we would add more conditional stuff at the toplevel that is easy to mess up.  I think it's better to keep the .pc files along with the library code that each .pc file is supposed to describe.

What about moving your test scripts to a common test directory instead?  The .pc files aren't specific to any particular build system (unlike cmake) so they don't need to live alongside the autotools test.

fghoussen:
I have never been happy with that neither but I'd like at least to keep things symmetric.

>  The makefiles would reference things outside of their directory

That's why at some point I moved all files (`*.pc`, `*.cmake`, `tst*.sh`) at the root level: you know where things are

> better to keep the .pc files along with the library code that each .pc file

There is 1 `*.cmake` file for both arpack and parpack, but, 2 `*.pc` file for each of them.
Do we want to split `*.cmake` in 2 to have one for arpack and another one for parpack?.... Is this worth?!...

turboencabulator:
> There is 1 `*.cmake` file for both arpack and parpack, but, 2 `*.pc` file for each of them. Do we want to split `*.cmake` in 2 to have one for arpack and another one for parpack?.... Is this worth?!...

I don't know enough about cmake to have an opinion on this.  For pkg-config the best practice seems to be having one .pc description file for each library/module.

fghoussen:
IMO, the repo should be symmetric (as arpack is!). `*.cmake` can be split: no time for that...

fghoussen:
@turboencabulator: can you rebase?... Sorry for the delay I forgot about this PR...

turboencabulator:
Sorry @fghoussen I forgot about this one as well.  It's now rebased.

CMakeLists.txt
```diff
@@ -847,15 +847,15 @@ foreach(lib ${LAPACK_LIBRARIES} ${MPI_Fortran_LIBRARIES})
 endforeach()
 string(REPLACE ";" " " PARPACK_PC_LIBS_PRIVATE "${PARPACK_PC_LIBS_PRIVATE}")
 
-configure_file(pkg-config/arpack.pc.in "${PROJECT_BINARY_DIR}/arpack${LIBSUFFIX}${ITF64SUFFIX}.pc" @ONLY)
-configure_file(pkg-config/parpack.pc.in "${PROJECT_BINARY_DIR}/parpack${LIBSUFFIX}${ITF64SUFFIX}.pc" @ONLY)
+configure_file(SRC/arpack.pc.in "${PROJECT_BINARY_DIR}/SRC/arpack${LIBSUFFIX}${ITF64SUFFIX}.pc" @ONLY)
+configure_file(PARPACK/SRC/MPI/parpack.pc.in "${PROJECT_BINARY_DIR}/PARPACK/SRC/MPI/parpack${LIBSUFFIX}${ITF64SUFFIX}.pc" @ONLY)
```
fghoussen: @turboencabulator Why did you move the pc files? (as the repo is maintained-only, I'd prefer minimal changes to avoid breaking everything!)

turboencabulator: It's documented in the commit message:  Each directory's Makefile is responsible for building/installing files in that directory.  If parpack isn't selected to be built, we skip over the parpack source directory and its .pc file won't be installed, and we don't need extra decisions elsewhere to skip installing that file.

This is also just moving the files back to their original locations (before 93310367025a01db9c582f2c2a264d8f91826626 moved them and caused this problem).

fghoussen:
@turboencabulator: thanks!

Issue #218, opened 2019-07-07
Arpack random generator cannot be seeded


pv:
## Expected behavior

It would nice to be able to set the random seed of Arpack's internal random number generator, for cases where more strict fp reproducibility is aimed at (needs of course also compiler and library support). There could be a new subroutine call in the API that sets it.

## Actual behavior

Currently, the rng seeding is done on the first call to each of *getv0.f routines, https://github.com/opencollab/arpack-ng/blob/c43cb868546e8f987cf19b3c2613c56be3865ee3/SRC/dgetv0.f#L202-L208

Specifying an initial vector is not sufficient, since new random vectors are sometimes needed along the run, https://github.com/opencollab/arpack-ng/blob/c43cb868546e8f987cf19b3c2613c56be3865ee3/SRC/dsaitr.f#L409 

## Where/how to reproduce the problem

- arpack-ng: as of 2019-07

fghoussen:
You may propose a PR. If so, think about also providing / updating iso_c_ binding API.

fghoussen:
BTW, not sure to get the pb right. Seeds are fixed: so generated random numbers should be the same if you provide the same initial vector, no ?

pv:
The seed is initialized only the first time `*getv0` is called in a process (save variables), so you'll get different results every time you invoke arpack in the same process. This means e.g. that libraries that use arpack cannot control the randomness, since they don't control the whole-program execution.

fghoussen:
OK, I can try to PR something about that.
So you need something to reset seed at every iteration of the RCI-while loop, right ?

pv:
I have a branch here: https://github.com/opencollab/arpack-ng/compare/master...pv:rng-seed

pv:
By the way, are the iso_c_bindings as they currently are done in arpack-ng correct for the `logical` type? Adding the `interface` declaration the compiler complains that `logical(kind=c_bool)` and `logical` are not compatible types, and indeed `false` seems to behave incorrectly if no conversion is done and interface is omitted?

fghoussen:
> I have a branch here: master...pv:rng-seed

Looks a bit fuzzy (many impacts) and incomplete (missing all p*getv0.f). Tried to PR some fix : not tested, @pv could you test if this behaves like you need ? (debug/stat common have already been initialized. I realized that, here, it is about save... Which may not behave _exactly_ like common). If this works, this should cover all cases.

[...]

PR #431, opened 2023-08-19
Simplify RNG seed initialization in PARPACK

FabienPean:
Following up #424, one can simplify the parpack seed initialization to fulfill the LAPACK requirements. Before, it was already a constant number which could handle "only" 1547 (0.5*(4095-1000-1)) proc id.

### DIFF
--- a/PARPACK/SRC/MPI/pcgetv0.f
+++ b/PARPACK/SRC/MPI/pcgetv0.f
@@ -177,7 +177,7 @@ c     | Local Scalars & Arrays |
 c     %------------------------%
 c
       logical    first, inits, orth
-      integer    idist, iseed(4), iter, msglvl, jj, myid, igen
+      integer    idist, iseed(4), iter, msglvl, jj, myid
       Real
      &           rnorm0
       Complex
@@ -214,32 +214,22 @@ c     | Executable Statements |
 c     %-----------------------%
 c
 c
+      if (inits) then
+c
 c     %-----------------------------------%
 c     | Initialize the seed of the LAPACK |
 c     | random number generator           |
+c     | Note: seed array elements must be |
+c     | between 0 and 4095 (12-bit coded) |
+c     | iseed(4) must be odd              |
 c     %-----------------------------------%
 c
-      if (inits) then
-c
-c        %-----------------------------------%
-c        | Generate a seed on each processor |
-c        | using process id (myid).          |
-c        | Note: the seed must be between 1  |
-c        | and 4095.  iseed(4) must be odd.  |
-c        %-----------------------------------%
-c
          call MPI_COMM_RANK(comm, myid, ierr)
-         igen = 1000 + 2*myid + 1
-         if (igen .gt. 4095) then
-            write(0,*) 'Error in p_getv0: seed exceeds 4095!'
-         end if
 c
-         iseed(1) = igen/1000
-         igen     = mod(igen,1000)
-         iseed(2) = igen/100
-         igen     = mod(igen,100)
-         iseed(3) = igen/10
-         iseed(4) = mod(igen,10)
+         iseed(1) = 1
+         iseed(2) = 3
+         iseed(3) = 5
+         iseed(4) = 1+2*mod(myid,2048)
 c
          inits = .false.
       end if
--- a/PARPACK/SRC/MPI/pdgetv0.f
+++ b/PARPACK/SRC/MPI/pdgetv0.f
@@ -178,7 +178,7 @@ c     | Local Scalars & Arrays |
 c     %------------------------%
 c
       logical    first, inits, orth
-      integer    idist, iseed(4), iter, msglvl, jj, myid, igen
+      integer    idist, iseed(4), iter, msglvl, jj, myid
       Double precision
      &           rnorm0, buf2(1)
       save       first, iseed, inits, iter, msglvl, orth, rnorm0
@@ -217,32 +217,22 @@ c     | Executable Statements |
 c     %-----------------------%
 c
 c
+      if (inits) then
+c
 c     %-----------------------------------%
 c     | Initialize the seed of the LAPACK |
 c     | random number generator           |
+c     | Note: seed array elements must be |
+c     | between 0 and 4095 (12-bit coded) |
+c     | iseed(4) must be odd              |
 c     %-----------------------------------%
 c
-      if (inits) then
-c
-c        %-----------------------------------%
-c        | Generate a seed on each processor |
-c        | using process id (myid).          |
-c        | Note: the seed must be between 1  |
-c        | and 4095.  iseed(4) must be odd.  |
-c        %-----------------------------------%
-c
          call MPI_COMM_RANK(comm, myid, ierr)
-         igen = 1000 + 2*myid + 1
-         if (igen .gt. 4095) then
-            write(0,*) 'Error in p_getv0: seed exceeds 4095!'
-         end if
 c
-         iseed(1) = igen/1000
-         igen     = mod(igen,1000)
-         iseed(2) = igen/100
-         igen     = mod(igen,100)
-         iseed(3) = igen/10
-         iseed(4) = mod(igen,10)
+         iseed(1) = 1
+         iseed(2) = 3
+         iseed(3) = 5
+         iseed(4) = 1+2*mod(myid,2048)
 c
          inits = .false.
       end if
--- a/PARPACK/SRC/MPI/psgetv0.f
+++ b/PARPACK/SRC/MPI/psgetv0.f
@@ -178,7 +178,7 @@ c     | Local Scalars & Arrays |
 c     %------------------------%
 c
       logical    first, inits, orth
-      integer    idist, iseed(4), iter, msglvl, jj, myid, igen
+      integer    idist, iseed(4), iter, msglvl, jj, myid
       Real
      &           rnorm0
       save       first, iseed, inits, iter, msglvl, orth, rnorm0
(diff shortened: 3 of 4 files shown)


fghoussen:
I'll have a closer look when possible.
@szhorvat:  Sounds right to you?

PARPACK/SRC/MPI/pdgetv0.f
```diff
@@ -204,7 +204,7 @@ subroutine pdgetv0
 c     | Intrinsic Functions |
 c     %---------------------%
 c
-      intrinsic    abs, sqrt
+      intrinsic    abs, sqrt, mod
```
szhorvat, line 207: Why is this change made only in this file and not the other three?

szhorvat:
Yes, this change makes sense to me.

fghoussen:
@FabienPean: BTW, why do you think this is better than the original way of seeding?
@pv: does it help #218?

PARPACK/SRC/MPI/pdgetv0.f
FabienPean: I started to add it until I saw in other files it was not necessary. Forgot to remove it, will do it.

FabienPean:
> @FabienPean: BTW, why do you think this is better than the original way of seeding? 

It is simpler, no error block if it reaches 1547 MPI proc, and does not make some obfuscating calculation `1000+2*myid+1` is still a constant seed per MPI proc. The only difference is how the value are spread over iseed. 
It was yielding
* 0<= iseed(1)<5
* 0<=iseed(2)<1000
* 0<=iseed(3)<100
* 1<iseed(4)<10
Since there is no recommendation on the spread of the initial seed from LAPACK, the new value values are as good as any. So I copied the initial part from the serial code for iseed(1-3) and used a modulo for iseed(4), which at least guarantee an initial seed uniqueness until 2048 MPI proc, then MPI processes 0 and 2048 will start to reuse the initial seed and therefore random values for their part of the problem.

In brief, this PR yields fewer LOC and reduced seed repetition for high MPI proc counts

fghoussen:
> It is simpler

I guess this is OK and will not trigger regression.

@caliarim: what do you think? Potential reg or fix related to https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/138?

fghoussen:
> related to https://github.com/JuliaLinearAlgebra/Arpack.jl/issues/138

Just realizing that this was serial (arpack) so parpack will have no impact here

szhorvat:
I am not an expert on RNGs and can't comment much, but here's a reason why this manner of seeding may not be ideal, depending on the type of RNG used: https://stats.stackexchange.com/a/233091/4764 I believe `myid` will increase sequentially from thread to thread.

PARPACK/SRC/MPI/pcgetv0.f
```diff
@@ -213,33 +213,23 @@ subroutine pcgetv0
...
-         iseed(1) = igen/1000
-         igen     = mod(igen,1000)
-         iseed(2) = igen/100
-         igen     = mod(igen,100)
-         iseed(3) = igen/10
-         iseed(4) = mod(igen,10)
+         iseed(1) = 1
+         iseed(2) = 3
+         iseed(3) = 5
+         iseed(4) = 1+2*mod(myid,2048)
```
fghoussen, line 232: https://github.com/opencollab/arpack-ng/pull/431#issuecomment-1698617728: Not sure what to do. Does this mean we should let `seed([123])` as is (or dependent of `rand`?), and, initialize RNG seed to `myid` (with some kind of `call random_seed(put=myid)`?) before changing `iseed(4)` to `iseed(4) = 1+2*mod(rand(),2048)`?

fghoussen:
> I believe myid will increase sequentially from thread to thread.

Yes, it will.

szhorvat:
Well, this may not be an issue at all here. Whether it is depends on how LAPACK's RNG works, and once again, I'm definitely not an RNG expert.

Also, problems would only occur if `iseed(4)` (and not `iseed(1)`) is the least significant part of the 48 bit value it operates with. I don't know of this is the case.

If we want to be concerned about this, we can make all of `iseed(1)`, `iseed(2)`, `iseed(3)` and `iseed(4)` dependent on `myid`.

But at this point, I'm really hoping to see 3.9.1 released as soon as possible :-)

FabienPean:
> I am not an expert on RNGs and can't comment much, but here's a reason why this manner of seeding may not be ideal, depending on the type of RNG used: https://stats.stackexchange.com/a/233091/4764 I believe `myid` will increase sequentially from thread to thread.

I am not entirely sure what you mean here. AFAIU, the link discusses the choice of an initial seed and seed sampling per thread across several run of an application. In the case here, the seed was already the same across runs. The risk is more about the lack of sensitivity of the RNG (small seed number variation yields a small spread in the number generated, i&approx;j &rArr; x&approx;y). A good characteristic of a RNG is to not have this problem, which I would expect form Lapack.

The documentation of Lapack does not mention advice or any restriction on the choice of seed ([dlaran](https://netlib.org/lapack/explore-html/d4/dff/dlaran_8f_a9890a3a41d07fe1f5656031eb668f806.html#a9890a3a41d07fe1f5656031eb668f806)) but diving deeper in the theory behind multiplicative congruential method might provide further insights (which I won't go into though)

 > This routine uses a multiplicative congruential method with modulus
  2**48 and multiplier 33952834046453 (see G.S.Fishman,
  'Multiplicative congruential random number generators with modulus
  2**b: an exhaustive analysis for b = 32 and a partial analysis for
  b = 48', Math. Comp. 189, pp 331-344, 1990).
>
> 48-bit integers are stored in 4 integer array elements with 12 bits
  per element. Hence the routine is portable across machines with
  integers of 32 bits or more.

Nevertheless, the change of this PR is mostly a cosmetic change, instead of spreading a seed across the 4 elements of the array in an arbitrary manner, it increases linearly in iseed(4). For a fully random seed at each run, I gave a try in #424 , but it made several tests to fail consistently, so I didn't pursue then because of the major change it would be and the lack of visibility on the reason for failure.

Comparison for a few MPI ranks:

| myid | Before | This PR |
|--------|--------|--------|
| 1 | [1,0,0,3] | [1,3,5,3] |
| 3 | [1,0,0,7] | [1,3,5,7] |
| 5 | [1,0,1,1] | [1,3,5,11] |

```fortran
         igen = 1000 + 2*myid + 1
         iseed(1) = igen/1000
         igen     = mod(igen,1000)
         iseed(2) = igen/100
         igen     = mod(igen,100)
         iseed(3) = igen/10
         iseed(4) = mod(igen,10)
```

szhorvat:
I experimented a bit with this specific RNG and it became clear that using sequential seeds is not acceptable.

>  A good characteristic of a RNG is to not have this problem, which I would expect form Lapack.

There is a misunderstanding here about what the goal of a PRNG is and what makes it good quality. For a given seed, a PRNG will produce a sequence of numbers, which should be uniformly distributed and should appear statistically independent.  This does not mean that if we plug in one seed to get the sequence $a_1, a_2, \dots$ and a second seed to get $b_1, b_2, \dots$ then $a_i$ and $b_i$ will be independent. In fact, it is a feature of some RNG algorithms that they _won't_ be independent if the seeds were chosen to be close.

Let's reproduce LAPACK's RNG algorithm. The `laruv` docs state,

```
  This routine uses a multiplicative congruential method with modulus
  2**48 and multiplier 33952834046453 (see G.S.Fishman,
  'Multiplicative congruential random number generators with modulus
  2**b: an exhaustive analysis for b = 32 and a partial analysis for
  b = 48', Math. Comp. 189, pp 331-344, 1990).
```

So we have
$$r_{i+1} = A r_i (\mod M)$$
where $M=2^{48}$ and $A=33952834046453$.

Note that with `laruv`, the "seed" is in fact the entire RNG state. Let's take consecutive seeds 1 and 3, and see what we get if we plot $(a_i, b_i)$.  Note that the state needs to be an odd number (the Fishman reference explain why).  I'll include the Mathematica code I used for this, for reference.

```
iter[k_] := Mod[33952834046453 k, 2^48]

a = NestList[iter, 1, 1000];
b = NestList[iter, 3, 1000];

ListPlot[Transpose[{a, b}], AspectRatio -> Automatic]
```

image

It's clear that there are strong correlations between the two sequences:

Let's try 1 and 11:

image

How about 1 and 1001?

image

Alright, now it's looking good.

Conclusion: The parallel threads should not be initialized with consecutive seeds.

> I am not entirely sure what you mean here. AFAIU, the link discusses the choice of an initial seed and seed sampling per thread across several run of an application.

The SE post discusses the same situation we have here. If we have 4 threads, is it okay to init them with consecutive seeds, for example 23, 24, 25, 26.  It states that with a Mersenne twister it is not, as the four sequences would not be independent.

szhorvat:
I'll put this here as well: https://stackoverflow.com/q/51184553/695132

fghoussen:
Interesting! So as stated here: https://stackoverflow.com/questions/51184553/do-consequtive-rng-seed-yield-independent-random-numbers, one way to go could be: "Take those sequential numbers and run them through a cryptographic hash, then use that as the seed. Completely repeatable yet uncorrelated."? Only for `iseed(4)`? Also for `iseed([123])`?

> It's clear that there are strong correlations

Never used Mathematica: not sure what the code does! You plot `a` on the x-axis and `b` on the y-axis, right?

szhorvat:
> "Take those sequential numbers and run them through a cryptographic hash, then use that as the seed. Completely repeatable yet uncorrelated."

That would be one possible option.

> Only for `iseed(4)`?

`seed` represents a 48-bit number split into four 12-bit parts. So, ideally, populate all of it with a hash function that produces at least 48 bits.

> Never used Mathematica

You don't need to read the code but I like to insert it in these situations for my own reference.

> You plot `a` on the x-axis and `b` on the y-axis, right?

Yes, up to the last plot.

The last plot shows the 100th random number in streams initialized with consecutive seeds, the point being again that consecutive seeds won't produce independent streams.

FabienPean:
Thanks that's clearer and they are compelling arguments. The preexisting code is maybe not optimal, but this PR would theoretically make it worse.

> > "Take those sequential numbers and run them through a cryptographic hash, then use that as the seed. Completely repeatable yet uncorrelated."
> 
> That would be one possible option.

Another option could be to run the `dlarnv` routine `myid` times from a common starting seed across all MPI processes, convert the generated doubles to suitable integers and use that as a seed for subsequent calls in parpack.

Issue #430, opened 2023-08-19
About pyarpack test suite


FabienPean:
The test `pyarpackSparseCGILU_tst ` regularly fails as noted in #424. Diving further into the pyarpack test suite and comparing it with arpackmm (which I believe they were built after?), I noticed a few discrepancies:

* `pyarpackSparseCGILU.py.in` tests the CG solver with ILU preconditioner on (1) complex double for **non-hermitian** standard problem, and (2) complex single **non-hermitian** **general** problem
* `pyarpackSparseBiCGDiag.py.in` tests the BiCG stabilized solver on (1) float double for **symmetric** standard problem, and (2) float single **symmetric** **general** problem
*  `pyarpackSparseLDLT.py.in` tests the LDL solver on (1) complex double for **non-hermitian** standard problem, and (2) complex single **non-hermitian** **general** problem
*  `pyarpackSparseLU.py.in` tests the LU solver on (1) float double for **symmetric** standard problem, and (2) float single **symmetric** **general** problem

1. CG solver is not guaranteed to solve correctly non-hermitian matrices (which can explain regular failures in #424). One could swap `pyarpackSparseBiCGDiag.py.in` and `pyarpackSparseCGILU.py.in`
2. LDL solver cannot handle non-hermitian problem normally (yet the tests still pass, dunno how Eigen handles it internally, perhaps the dissymmetry is not large enough), so perhaps `pyarpackSparseLU.py.in` and `pyarpackSparseLDLT.py.in` should be swapped
3. Some of pyarpack solves the general problem while arpackmm test suite does not solve them at all, I believe it is more of an issue with arpackmm (reported in #429)

fghoussen:
Sounds right! Any help to improve is appreciated :)

> I believe they were built after?

Yes!

fghoussen:
Does #432 fixes everything?

FabienPean:
I had yet to get the full picture, even if A and B are hermitian in the general problem, the operation (OP in the Fortran code) is not guaranteed to keep the properties required by the solver in mode &geq; 3 because of the shift 😐. I made some changes in #436 but those changes are as good as any, as long as CG and LDL solve for symmetric problem.

fghoussen:
Fixed by #436

PR #432, opened 2023-08-20
Fix pyarpack

fghoussen:
## Pull request purpose

fixing issue #430

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -118,9 +118,9 @@ jobs:
       - name: Build boost-python for python3 (not provided by apt-cache)
         run : |
           sudo apt-get -y install wget
-          wget https://sourceforge.net/projects/boost/files/boost/1.79.0/boost_1_79_0.tar.gz
-          tar -xf boost_1_79_0.tar.gz
-          cd boost_1_79_0
+          wget https://sourceforge.net/projects/boost/files/boost/1.83.0/boost_1_83_0.tar.gz
+          tar -xf boost_1_83_0.tar.gz
+          cd boost_1_83_0
           ./bootstrap.sh --with-libraries=python --with-python=/usr/bin/python3 --with-toolset=gcc
           sudo ./b2 toolset=gcc install
           sudo apt-get install locate
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ arpack-ng - 3.9.1
  * README: Add details on Windows installation.
 
 [ Franck Houssen ]
+ * [BUG FIX] Fix pyarpack samples
  * [BUG FIX] Fix install: headers in /path/to/local/include/arpack for ICB samples
  * [BUG FIX] Fix install: headers in /path/to/local/include/arpack
  * arpackmm: allow for using LA/SA magnitudes.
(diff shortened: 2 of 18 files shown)


FabienPean:
I don't think all those pyarpack changes are warranted. By changing all input matrices to an hermitian/symmetric matrix, it reduces the coverage of solvers which handle non-symmetric (LU, QR, etc).

fghoussen:
> I don't think all those pyarpack changes are warranted.

They at least make sense! As you remarked it, hermitian makes better sense for complex. 

> By changing all input matrices to an hermitian/symmetric matrix, it reduces the coverage of solvers which handle non-symmetric (LU, QR, etc).

The coverage is not that bad (about 70% AFAIR - we had one when CI was done with travisCI). Historically, the coverage is the one of the original arpack repo: no dev from there (`arpack-ng` is a fork that "only" maintain the original arpack) so coverage has (and will) always be the same. These tests were only added to make sure that stuffs like ICB, arpackSolver.hpp and python binding (based on ICB) were working (not considering coverage) so solving the same case is OK if it helps stability/repeatability.

FabienPean:
> The coverage is not that bad (about 70% AFAIR - we had one when CI was done with travisCI). Historically, the coverage is the one of the original arpack repo: no dev from there (`arpack-ng` is a fork that "only" maintain the original arpack) so coverage has (and will) always be the same. These tests were only added to make sure that stuffs like ICB, arpackSolver.hpp and python binding (based on ICB) were working (not considering coverage) so solving the same case is OK if it helps stability/repeatability.

True, the complex problem anyway redirect to the same Fortran functions under the hood (cnaupd/cneupd), there is no symmetric variants of those (csaupd/cseupd) so using hermitian is the same as non-hermitian. If arpackmm tests non-symmetric problem and solver, so pyarpack could check whatever.

By browsing `cnaupd` documentation, I noticed a detail of importance:

> This is intended to be used to find a few eigenpairs of a complex linear operator OP with respect to a semi-inner product defined by a **hermitian positive semi-definite real matrix B**. B may be the identity matrix.  NOTE: if both OP and B are real, then ssaupd or snaupd should be used.

It seems to imply that B must be real (no imaginary part) and hermitian (so symmetric?) semi-definite matrix. The current B matrices used for the tests might be not valid ?

fghoussen:
>  complex problem anyway redirect to the same Fortran functions under the hood (cnaupd/cneupd), there is no symmetric variants of those (csaupd/cseupd) so using hermitian is the same as non-hermitian

Right!

> The current B matrices used for the tests might be not valid ?

Main goal to achieve here is to get stable/repeatable tests (no focus on coverage): is it worth to make B valid/real in #436 to improve test stability?

fghoussen:
#436 fix is prefered.

FabienPean:
> Main goal to achieve here is to get stable/repeatable tests (no focus on coverage): is it worth to make B valid/real in #436 to improve test stability?

Hard to say, I just saw that in the comments for cnaupd, but the exact description (real matrix) is not repeated when describing the modes 2 and 3 below. Besides, the tests with complex values in Fortran (in `EXAMPLES/COMPLEX`) never involve an imaginary part for the matrices. Since the test was passing already and without clear idea on how to improve, I will avoid fiddling further :)

PR #433, opened 2023-08-20
Fix arpackmm test suite

### LINKED ISSUE #429: About arpackmm test suite
Diving further into the different pyarpack examples (due to #424), I came to check those generated for [arpackmm](https://github.com/opencollab/arpack-ng/blob/master/EXAMPLES/MATRIX_MARKET/arpackmm.sh), one can observe a few discrepancies.

In [arpackmm.sh](https://github.com/opencollab/arpack-ng/blob/master/EXAMPLES/MATRIX_MARKET/arpackmm.sh):

https://github.com/opencollab/arpack-ng/blob/e7cf106bc204ed7f7909d906fd9b1c3d18e86f76/EXAMPLES/MATRIX_MARKET/arpackmm.sh#L54-L59

1. The first regex removes non symmetric problem to be solved by the (stabilized) BiCG solver, which can handle them. (See https://eigen.tuxfamily.org/dox/group__TopicSparseSystems.html)

https://github.com/opencollab/arpack-ng/blob/e7cf106bc204ed7f7909d906fd9b1c3d18e86f76/EXAMPLES/MATRIX_MARKET/arpackmm.sh#L18-L19

2. Besides the discrepancy between decomposition solvers and iterative solvers in the first snippet, general problem are always skipped anyway, which seems an unlikely desired behavior. Especially since some of those are already solved with pyarpack (#430). 

https://github.com/opencollab/arpack-ng/blob/e7cf106bc204ed7f7909d906fd9b1c3d18e86f76/EXAMPLES/MATRIX_MARKET/arpackmm.sh#L60-L64

3. In particular the general problem always involves symmetric SPD/PD matrices so LDL and LL can work too, as used in pyarpack test suite (`pyarpack*LLT.py.in`, `pyarpack*LDLT.py.in`)

FabienPean:
Fix #429

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ arpack-ng - 3.9.1
 
 [ Fabien Péan ]
  * [BUG FIX] Tests for PARPACK with C/C++ bindings icb_parpack_c and icb_parpack_cpp are now really parallel and split the problem across MPI processes.
+ * Update arpackmm test suite: enable solving non-symmetric tests with BiCG solver 
 
 [ Szabolcs Horvát ]
  * [BUG FIX] Ensure that LAPACK RNG state is propagated (regression in 3.9.0).
--- a/EXAMPLES/MATRIX_MARKET/arpackmm.sh
+++ b/EXAMPLES/MATRIX_MARKET/arpackmm.sh
@@ -51,7 +51,7 @@ do
                       export extraGenPb="$shiftOpt" # Force shift if genPb.
                     fi
 
-                    if [[ "$slv" == *CG* ]]; then
+                    if [[ "$slv" == *" CG "* ]]; then
                       if [[ "$eigPb" == *nonSymPb* ]]; then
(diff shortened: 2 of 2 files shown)


FabienPean:
Welp, at least I found out why the arpackmm tests were diverging for jobs `ubuntu_latest_autotools` and `ubuntu_latest_cmake`

fghoussen:
Seems there is no clue in traces.... I'll have a look when possible.

FabienPean:
The general part seems quite problematic. I will split this PR and just enable BiCG for non symmetric problems. The general problem should be handle seprately as I don't think I will be able to dive in further

fghoussen:
OK, good to merge? Still something you like to add?

FabienPean:
That's it, ready to be merged. The general eigenproblem is generally more complex than anticipated (🤡 pun intented) and I will likely leave it untouched.

fghoussen:
thanks!

PR #435, opened 2023-08-28
Update README.md MKL options

VinceNeede:
cmake option for mkl BLA_VENDOR=INTEL works for obsolete verdions, replaced with INTEL10_64lp and INTEL10_64ilp for lp64 and ilp64 respectively, as reported here: https://cmake.org/cmake/help/latest/module/FindBLAS.html


fghoussen:
Thanks!

PR #437, opened 2023-09-20
Update RNG seed initialization in PARPACK

FabienPean:
Following-up and trying to wrap-up #431. Taking into account the elements raised by https://github.com/opencollab/arpack-ng/pull/431#issuecomment-1705620924 using basic tools.

Does this change seems appropriate to you @szhorvat @fghoussen ?

Each seed is initialized from a random generator coming from a master seed by running `myid` times the RNG. For a given seed, successive call to the generator should produce an independent sequence according the info, then converting those number to fit within the LAPACK seed constraints.

### DIFF
--- a/PARPACK/SRC/MPI/pcgetv0.f
+++ b/PARPACK/SRC/MPI/pcgetv0.f
@@ -177,9 +177,9 @@ c     | Local Scalars & Arrays |
 c     %------------------------%
 c
       logical    first, inits, orth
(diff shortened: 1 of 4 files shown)


szhorvat:
I'm not completely comfortable with this and I'll try to get back to you later.

One (not the only) concern here is whether the process ID can have large values, causing a long iteration.

sylvestre:
is there a way to test this? thanks

PR #438, opened 2023-10-14
configure should still install cmake files

sylvestre:


mmuetzel:
The CMake config files cannot be used on their own. They include the following line:
https://github.com/opencollab/arpack-ng/blob/947a004c85e8c1f93ed065e5211c3697cbd75689/cmake/arpackng-config.cmake.in#L35

The file `arpackngTargets.cmake` is generated by CMake. It isn't generated by the Makefile rules (nor by the configure script).
That leads to the odd situation that trying to use that config file in downstream CMake projects leads to the following fatal error:
```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/arpack-ng/arpackng-config.cmake:35 (include):
  include could not find requested file:
```

CMake doesn't support catching fatal errors and continuing execution.

On the other hand, if no CMake config files would be installed at all, CMake determines that gracefully in downstream projects, and they can fall back to alternative ways to locating ARPACK-ng.

See, e.g., https://github.com/ElmerCSC/elmerfem/issues/907 where this came up with the version of ARPACK-ng that is distributed by Debian. That version is apparently built using the Makefile rules.

Would it be ok to revert this change?

mmuetzel:
I opened #492 that reverts this PR.

PR #441, opened 2023-11-13
configure.ac: fix bashisms

thesamesam:
## Pull request purpose

configure scripts need to be runnable with a POSIX-compliant /bin/sh.

On many (but not all!) systems, /bin/sh is provided by Bash, so errors like this aren't spotted. Notably Debian defaults to /bin/sh provided by dash which doesn't tolerate such bashisms as '=='.

This retains compatibility with bash.

Fixes configure warnings/errors like:
```
checking for cheev_ in -llapack... yes
./configure: 8590: test: x: unexpected operator
checking how to print strings... printf
```

- Remove bashisms from the configure script, as required by autotools.


fghoussen:
Thanks!

thesamesam:
Thank you!

PR #444, opened 2023-11-22
Fix examples BAND _nbdr2 returning error info = -9

wo80:
## Pull request purpose

Fix examples BAND/cnbdr2.f and BAND/znbdr2.f returning error info = -9, meaning the initial residual was 0.

Though a standard eigenvalue problem is solved, both examples access the M matrix. But that is never initialized, so this PR sets M to be the identity matrix.

See #443 for discussion.


fghoussen:
Thanks!

Issue #448, opened 2024-01-27
libparpack built without using FFLAGS when using autotools


loqs:
## Expected behavior

FFLAGS will be used for all parts of the build.

## Actual behavior

FFLAGS are used for everything except libparpack.

## Where/how to reproduce the problem

- arpack-ng: release or commit (SHA)
3.9.1 40329031ae8deb7c1e26baf8353fa384fc37c251 
- OS: debian ? ubuntu ? mac ?...
Arch Linux
- compiler: gcc ? intel ? clang ? version ?...
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --enable-languages=ada,c,c++,d,fortran,go,lto,objc,obj-c++ --enable-bootstrap --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --with-build-config=bootstrap-lto --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-libstdcxx-backtrace --enable-link-serialization=1 --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-werror
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 13.2.1 20230801 (GCC) 
$ f90 -v
Using built-in specs.
COLLECT_GCC=f95
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
[...]

PR #449, opened 2024-01-27
Use FFLAGS for PARPACK with autotools and mpi

loqs:
Fixes https://github.com/opencollab/arpack-ng/issues/448.

### DIFF
--- a/PARPACK/SRC/MPI/Makefile.am
+++ b/PARPACK/SRC/MPI/Makefile.am
@@ -1,8 +1,6 @@
 AUTOMAKE_OPTIONS = subdir-objects # Needed as debug/stat* are not in current directory.
 
 F77 = $(MPIF77)
-FFLAGS_SAV = @FFLAGS@
-FFLAGS =
 
 PSRC = pcontext.f
 SSRC = psnaitr.f psnapps.f psnaup2.f psnaupd.f psneigh.f psneupd.f psngets.f \


PARPACK/SRC/MPI/Makefile.am
```diff
@@ -1,8 +1,6 @@
 AUTOMAKE_OPTIONS = subdir-objects # Needed as debug/stat* are not in current directory.
 
 F77 = $(MPIF77)
-FFLAGS_SAV = @FFLAGS@
```
fghoussen, line 4: @sylvestre: OK for you? Was this intentional?

loqs: Yes this was intentional.  I missed it is still used by `libparpack@LIBSUFFIX@@ITF64SUFFIX@_la_FFLAGS = $(FFLAGS_SAV)` which can now use FFLAGS instead.  Or keep it to a one line change?

fghoussen: > libparpack@LIBSUFFIX@@ITF64SUFFIX@_la_FFLAGS = $(FFLAGS_SAV) which can now use FFLAGS instead. 

I guess this is a good move.

> Or keep it to a one line change?

@sylvestre: what do you think? Was this `FFLAGS_SAV` intentional (bug fix?) back in the old days?!

sylvestre: I don't know, is there anything interesting in the git blame/log ? :)

fghoussen: @turboencabulator:  git-blame is on you! :)
Why did you need `FFLAGS_SAV`? Bug fix? If yes, which one? Can you review this PR?

turboencabulator: I just moved those lines, they've been there since the beginning.

But I do know why they exist:  Automake sets the default FFLAGS to `-g -O2` which conflicts with the two files we're trying to build without optimizations.  Clearing FFLAGS is a dirty hack to avoid this conflict.  Without it, Makefile looks like this:
```
FFLAGS = -g -O2
...
libparpack_noopt_la_FFLAGS = -O0
...
libparpack_noopt_la-pslamch10.lo: pslamch10.f
        $(AM_V_F77)$(LIBTOOL) $(AM_V_lt) --tag=F77 $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) --mode=compile $(F77) $(libparpack_noopt_la_FFLAGS) $(FFLAGS) -c -o libparpack_noopt_la-pslamch10.lo `test -f 'pslamch10.f' || echo '$(srcdir)/'`pslamch10.f

libparpack_noopt_la-pdlamch10.lo: pdlamch10.f
        $(AM_V_F77)$(LIBTOOL) $(AM_V_lt) --tag=F77 $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) --mode=compile $(F77) $(libparpack_noopt_la_FFLAGS) $(FFLAGS) -c -o libparpack_noopt_la-pdlamch10.lo `test -f 'pdlamch10.f' || echo '$(srcdir)/'`pdlamch10.f
```
Since `$(FFLAGS)` comes after `$(libparpack_noopt_la_FFLAGS)`, the `-O2` is last and wins.  So this will break the no-optimization stuff.

Two more questions:
1. What problem was the no-optimization stuff trying to fix?  Do we still need these two files to be built unoptimized?  If not, then let's delete libparpack_noopt.la, build these two files as part of libparpack.la, and delete everything related to FFLAGS in this file.
2. What `FFLAGS` are you trying to add?  Is it something that `configure` should be detecting and adding to `AM_FFLAGS`?

turboencabulator: I mentioned `AM_FFLAGS` in my previous comment but I just realized that those are getting ignored too.  That might also be part of the problem.

fghoussen: > What problem was the no-optimization stuff trying to fix?

Don't know!
@loqs and @turboencabulator: I let you converge to the best solution to PR. Let me know when the PR is ready to be merged.

fghoussen: @loqs: is there the same problem with cmake? I guess no.
Otherwise feel free to PR the equivalent fix at cmake side.

loqs: >     2. What `FFLAGS` are you trying to add?  Is it something that `configure` should be detecting and adding to `AM_FFLAGS`?
`FFLAGS=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -fstack-clash-protection -fcf-protection  -ffile-prefix-map=/build/arpack/src=/usr/src/debug/arpack`
I detected -fcf-protection was missing from PARPACK.

loqs: > @loqs: is there the same problem with cmake? I guess no. Otherwise feel free to PR the equivalent fix at cmake side.
autotools only.  cmake is building pdlamch10.f and pslamch10.f with optimization
https://github.com/opencollab/arpack-ng/blob/ba440d13229972b161de37d08dd61bd5a0aa07f2/CMakeLists.txt#L385

turboencabulator: I was thinking your FFLAGS might be related to MPI, but doesn't look like it.

Usually when I see `-O0` it's because the code has some obscure bug that the optimizer exposes.  But these two files seem fairly simple, and if CMake is building them optimized and nobody has complained yet... I say we remove the no-optimization stuff.

loqs:
Closing in favor of #450.

PR #415, opened 2023-04-15
CI: Setup workflow for MinGW-w64 CLANG64

MehdiChinoune:
## Pull request purpose

This setup CI workflow for MSYS2/CLANG64
It completes the work done in #412, The workflow was failing, but we fixed that issue.

## Detailed changes proposed in this pull request

- Setup CI workflow for MinGW-w64 CLANG64
- Test with MPI=ON/OFF
- Test with INTERFACE64=ON/OFF

PR #459, opened 2024-06-21
CI: Add CI for CLANG64 environment of MSYS2

mmuetzel:
## Pull request purpose

Add CI for CLANG64 environment of MSYS2.
Add runners for different configurations (MPI, INTERFACE64) on MinGW-w64.

This picks up on #415.

The [CLANG64 environment](https://www.msys2.org/docs/environments/) of MSYS2 uses a toolchain that it almost entirely based on LLVM tools. That includes compilers (incl. LLVM Flang), linker, and LLVM libc++ (instead of GNU's libstdc++). Most common build environments are based on GNU tools (apart from maybe macOS and MSVC on Windows).
It might be worth testing unusual build environments in CI to make sure they don't break unintentionally.

This PR adds runners for that build environment.
It also adds Eigen3 as an additional build dependency on MinGW to test it on that platform.
Additionally, it adds runners for the following configurations (to the UCRT64 and CLANG64 environments):
* MPI=ON/OFF
* INTERFACE64=ON/OFF

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -324,10 +324,29 @@ jobs:
           cmake --build . -v
           echo "::endgroup::"
       - name: Run tests
+        id: run-ctest
         run: |
           if [[ ${{ matrix.mpi }} == ON ]]; then
             export PATH="/c/Program Files/Microsoft MPI/Bin":$PATH # add mpiexec to msys2 path
           fi
           export PATH="${GITHUB_WORKSPACE}/build":$PATH # add libarpack.dll to msys2 path for tests that run in different directory
           cd build
-          ctest --output-on-failure
+          ctest
+      - name: Re-run tests
+        if: always() && (steps.run-ctest.outcome == 'failure')
+        timeout-minutes: 60
+        run: |
+          if [[ ${{ matrix.mpi }} == ON ]]; then
+            export PATH="/c/Program Files/Microsoft MPI/Bin":$PATH # add mpiexec to msys2 path
+          fi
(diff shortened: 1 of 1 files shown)


mmuetzel:
Adding the path to the built libarpack.dll to the `PATH` variable is just to work around an issue that might better be fixed in the CMake build system.
I'd be happy to look into proposing a change for that. But it would be nice if this could be merged first to make it easier to check if any follow-up change actually works.

mmuetzel:
Is there an obvious reason why `arpackmm_tst` takes longer on Windows than on, e.g., Ubuntu?
ubuntu_latest_cmake:
```
      Start 91: arpackmm_tst
91/95 Test #91: arpackmm_tst .....................   Passed   43.28 sec
```
MinGW-w64 UCRT64 INTERFACE64=OFF/MPI=OFF
```
      Start 82: arpackmm_tst
82/84 Test #82: arpackmm_tst .....................   Passed  148.70 sec
```

fghoussen:
Could you fix the test that fails?

> Is there an obvious reason why arpackmm_tst takes longer on Windows than on, e.g., Ubuntu?

Nope

mmuetzel:
> Could you fix the test that fails?

The macOS one?
I don't have local access to macOS. But I can try...

fghoussen:
> The macOS one?

Yep. No mac too here. I hope `pip3 install numpy>=2.0.0`could be enough...

mmuetzel:
> > The macOS one?
> 
> Yep. No mac too here. I hope `pip3 install numpy>=2.0.0`could be enough...

I've opened #460 that should fix the version conflict on that runner.

PR #460, opened 2024-06-22
CI: Install numpy from Homebrew on macOS

mmuetzel:
## Pull request purpose

Avoid version conflict with numpy on macOS runners.

Binary packages from different distributors could be built with different versions of dependencies. If these different versions are using an incompatible API or ABI this could lead to errors on build or execution time.

Try to install as many packages as possible from the same distributor (Homebrew in this case). I.e., no longer install numpy from pip while Python is installed from Homebrew.
This also simplifies the build rules for that job a bit. No virtual Python environment is needed.

Also, use non-conflicting configuration settings when it comes to the BLAS and LAPACK implementations on the macOS runners.


fghoussen:
@mmuetzel: thanks for your help and contribution! :)

PR #461, opened 2024-06-22
Make ctests platform independent

mmuetzel:
## Pull request purpose

Change configuration of some ctests on Windows so they work "out of the box".

Windows doesn't have a mechanism similar to rpaths. Instead, shared libraries are searched in the current working directory and in the directory with the executable followed by directories in the environment variable PATH.

Add the path to the shared libarpack library to PATH for ctests that run executables that are located in a different directory and that set the current working directory not to the directory with the libarpack.dll. Use a generator expression that should be working independent on the CMake generator.

Also remove the previously added work-around from the CI rules for MinGW that is no longer needed with that change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -711,6 +711,11 @@ function(build_tests)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.sh)
       add_test(NAME issue215_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} issue215.sh)
+      if (WIN32 AND BUILD_SHARED_LIBS)
+        set_tests_properties(arpackmm_tst issue401_tst issue215_tst
+          PROPERTIES
+          ENVIRONMENT_MODIFICATION "PATH=path_list_prepend:$<TARGET_FILE_DIR:arpack>")
+      endif()
     endif()
 
     if (PYTHON3)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -626,7 +626,23 @@ endif()
 ############################
 # TEST
 ############################
-function(build_tests)  
+
+function(add_test_with_rpath test_name)
+  # Add test which sets a current working directory that doesn't contain the
+  # libarpack library.
+  # Windows doesn't have a mechanism similar to RPATH on Linux or macOS.
+  # On Windows, if a test is running in a working directory that doesn't contain
+  # the ARPACK (and PARPACK) libraries, the path to these libraries has to be
+  # added to the PATH environment variable.
+  add_test(NAME ${test_name} ${ARGN})
+  if (WIN32 AND BUILD_SHARED_LIBS)
+    set_tests_properties(${test_name}
+      PROPERTIES
+      ENVIRONMENT_MODIFICATION "PATH=path_list_prepend:$<TARGET_FILE_DIR:arpack>")
+  endif()
+endfunction()
+
+function(build_tests)
   set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/TESTS)
 
   add_executable(dnsimp_test TESTS/dnsimp.f TESTS/mmio.f TESTS/debug.h)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -711,6 +711,11 @@ function(build_tests)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.sh)
       add_test(NAME issue215_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} -eu issue215.sh)
+      if (WIN32 AND BUILD_SHARED_LIBS)
+        set_tests_properties(arpackmm_tst issue401_tst issue215_tst
+          PROPERTIES
+          ENVIRONMENT_MODIFICATION "PATH=path_list_prepend:$<TARGET_FILE_DIR:arpack>")
```
fghoussen, line 717: Shouldn't be this done for all tests? Why only theses 3 ones?

mmuetzel: Good point. I forgot to mention the crucial point in the PR description: `.dll` are also loaded from the current working directory on Windows (a security nightmare!).

Afaict, these three tests are the only ones that set the working directory to somewhere other than the directory with the libarpack.dll.

I'll update the description of the PR accordingly.

fghoussen: > these three tests are the only ones that set the working directory to somewhere other than

These 3 tests seem to be the only ones to be ran from a sh script with modified WORKING_DIRECTORY: https://github.com/opencollab/arpack-ng/blob/77ba50e8007695af4299dc9709da709e48288eca/CMakeLists.txt#L706-L713

May be changing `add_test` or/and `configure_file` could be enough?

Does autotools have this problem (we try as much as possible to make the 2 cmake/autotools work the same way)?

fghoussen: > 3 tests seem to be the only ones to be ran from a sh script with modified WORKING_DIRECTORY

WORKING_DIRECTORY could maybe simply be suppressed from `add_test` for these 3 tests: did you try? This way there would be nothing specific to any tests

mmuetzel: A possible alternative could be to put everything for these tests in the `CMAKE_BINARY_DIR` (i.e., the directory with the built libarpack). That is at least the generated .mtx files and the arpackmm binary. (The generated shell script could probably stay in a dedicated directory.)
The shell script tries to run the arpackmm binary in the current working directory. And iiuc, the arpackmm binary is looking for the .mtx files in the current working directory, too.

But that seems more messy than the current setup imho. Currently, the files for the Eigen tests are nicely separated in their own directory. Please, let me know if you prefer to change that anyway.

I haven't checked running the tests with the autotools build system on Windows yet. I'll try that after we know *how* this issue should be resolved.

fghoussen: > A possible alternative could be to put everything for these tests in the CMAKE_BINARY_DIR

Not sure what to do here. Both solution look similar to me. @sylvestre: what do you say?

CMakeLists.txt
```diff
@@ -711,6 +711,11 @@ function(build_tests)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/issue215.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue215.sh)
       add_test(NAME issue215_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} -eu issue215.sh)
+      if (WIN32 AND BUILD_SHARED_LIBS)
+        set_tests_properties(arpackmm_tst issue401_tst issue215_tst
```
fghoussen, line 715: Could you add a cmake `function(add_test_with_working_dir test_name)` to gather `add_test` and `set_tests_properties` for the 3 tests that need a working dir? Would be clearer I guess

mmuetzel: I added that function in the last commit.

Is that what you had in mind?

fghoussen: > Is that what you had in mind?

Yes!

CMakeLists.txt
```diff
@@ -704,13 +716,19 @@ function(build_tests)
       configure_file(EXAMPLES/MATRIX_MARKET/B.mtx  ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/B.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/Bz.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Bz.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/arpackmm.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/arpackmm.sh)
-      add_test(NAME arpackmm_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} -eu arpackmm.sh)
+      add_test_with_working_dir(arpackmm_tst
+        WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
```
fghoussen, line 720: Can you move the `WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` and the `COMMAND` inside the `add_test_with_working_dir` function

mmuetzel: That would make that function less versatile.
Let's say a future test would require the working directory to be set to somewhere in the source tree. Then, the new function wouldn't work.

I also don't know why you mean by moving the `COMMAND` to the new function. It is different for each of the three tests.

CMakeLists.txt
```diff
@@ -626,7 +626,19 @@ endif()
 ############################
 # TEST
 ############################
-function(build_tests)  
+
+function(add_test_with_working_dir test_name)
```
fghoussen, line 630: As is the function name does not do what it says

I though at something like:

```
function(add_test_with_working_dir test_name test_dir test_cli)
  add_test(NAME ${test_name} WORKING_DIRECTORY ${test_dir} COMMAND {test_cli})
```

mmuetzel: There is a reason why CMake recommends to use the function syntax with options instead of fixed arguments. See, e.g., at the bottom of the following page:
https://cmake.org/cmake/help/latest/command/add_test.html
> The command syntax above is recommended over the older, less flexible form

Imho, the changes you are proposing would be a pessimization to the currently proposed implementation (mainly because it would be less flexible).

Please, let me know if this is a show-stopper for you. Otherwise, I'd prefer to not do the change that you are proposing.

As far as the function name is concerned: The new function should be used if tests are added that manually change the working directory to somewhere that doesn't contain the libarpack library. So, the current name isn't misleading imho.
Please, let me know if you prefer a different name for that function.

mmuetzel: Sorry to bother.
Any feedback on the last comment?

fghoussen: > As is the function name does not do what it says

Can you pick a function name what do what it says? `add_test_with_custom_rpath` or similar and explain/comment what you do there "on windows, add path to lib in PATH as rpath is not supported"

CMakeLists.txt
```diff
@@ -704,13 +716,19 @@ function(build_tests)
...
       configure_file(EXAMPLES/MATRIX_MARKET/Bz.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Bz.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/arpackmm.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/arpackmm.sh)
-      add_test(NAME arpackmm_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} -eu arpackmm.sh)
+      add_test_with_working_dir(arpackmm_tst
+        WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+        COMMAND ${BASH_PROGRAM} -eu arpackmm.sh)
       configure_file(EXAMPLES/MATRIX_MARKET/issue401.mtx ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue401.mtx)
       configure_file(EXAMPLES/MATRIX_MARKET/issue401.sh ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/issue401.sh)
-      add_test(NAME issue401_tst WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} COMMAND ${BASH_PROGRAM} -eu issue401.sh)
+      add_test_with_working_dir(issue401_tst
```
fghoussen, line 724: What is the change here? Stick to minimal change please

CMakeLists.txt
```diff
@@ -795,7 +813,7 @@ endfunction(build_tests)
 
 if(TESTS)
     enable_testing()
-    set(CMAKE_CTEST_COMMAND ctest -V)   
+    set(CMAKE_CTEST_COMMAND ctest -V)
```
fghoussen, line 816: OK to remove extra spaces

mmuetzel:
I hope the latest change addresses your comments.

fghoussen:
> I hope the latest change addresses your comments.

Yes, thanks!

fghoussen:
Can you update the changelog?

mmuetzel:
Is that Changelog entry ok?

mmuetzel:
@sylvestre: Thank you for merging.

sylvestre:
thank you for you contrib :)

Issue #465, opened 2024-08-29
Program crash when calling DNEUPD after exiting DNAUPD with info=0 and iparam(5) > NEV


JarRosenberg:
Hello, I've run into what I think is a bug using the `DNAUPD` and `DNEUPD` functions in shift-invert mode to calculate eigenvalues of some admittedly dificult matrices. The short of it is that sometimes calling `DNEUPD` after `DNAUPD` leads to a program crash, despite the lack of an error on return from `DNAUPD`. I've found the when this occurs, `iparam(5)`, the number of converged eigenvalues, is greater than `NEV`, the number of requested eigenvalues. I have attached a txt containing a sample matrix ([failed_mat.txt](https://github.com/user-attachments/files/16792292/failed_mat.txt)), and a simple program that triggers the crash. More details provided below.

## Expected behavior

The `info` variable would be non-zero after calling `DNAUPD`, indicating an error has occured, when trying to solve the provided matrix and others like it. Alternatively, calling `DNEUPD` would return the correct eigenvalue.

## Actual behavior

Calling `DNEUPD` after a "sucessful" exit from `DNAUPD` causes the program to crash with the error shown below solving the attached matrix.

## Where/how to reproduce the problem

- arpack-ng: current release
- OS: macOS Sonoma 14.5
- compiler: gfortran version 13.2.0
- environment: FFLAGS = -std=f2018 -finit-real=snan -fbacktrace -fmax-errors=25 -fPIC -ffpe-trap=invalid,overflow,zero -fopenmp --Og -ggdb -fcheck=all -Wall -Wno-unused-dummy-argument -Wno-maybe-uninitialized -finline-limit=0
- configure: Unkown, though likely just the minimum settings needed to run on a Mac with an M1 chop.
- input data: See the attached file `failed_mat.txt` for the matrix in question.

## Steps to reproduce the problem

- Run the following program, attempting to solve the provided matrix ([failed_mat.txt](https://github.com/user-attachments/files/16792292/failed_mat.txt)) in shift-invert mode for a few eigenvalues with the largest magnitude, in the same directory as `failed_mat.txt`. Note that the calls to ARPACK functions are in the subroutine `eigen_shift_invert`.

``` FORTRAN
program arpack_test

   use ISO_FORTRAN_ENV
   use f95_lapack

   implicit none

   integer, parameter :: RD = REAL64
   integer, parameter :: sys_size = 42
[...]

PR #468, opened 2024-09-14
Fix arpack solver

fghoussen:
## Pull request purpose

Fixing limitations of arpack solver: in connection with #465


fghoussen:
@sylvestre: long story short, I had to change the API of the arpack solver and python binding based on it. New release?

sylvestre:
>  I had to change the API of the arpack solver and python binding based on it.

why ? 
just in the C++ or the fortran too?

fghoussen:
> why ?

To fix problems related to the way tolerances are used.

> just in the C++ or the fortran too?

C++ et python only. N ot C neither fortran.

PR #478, opened 2025-10-03
Reorder library linking for robustness

TLCFEM:
When `lapack` and `blas` are provided as static libraries, the linking order matters.

The current order links `blas` first, resulting in unresolved references in `lapack`.

Swap linking order would make it more robustness.


fghoussen:
> When lapack and blas are provided as static libraries, the linking order matters.
> The current order links blas first, resulting in unresolved references in lapack

Indeed, sounds right!

Can you factorize 'LAPACK::LAPACK BLAS::BLAS' in a variable (like LAPACK_BLAS?) and use it in each place where LAPACK_BLAS is needed

Also can you make the sample change in m4 autotools macro: use of ax_lapack ax_blas

Thanks

fghoussen:
Good to merge for me when CI passed. @sylvestre: any more remarks?

PR #484, opened 2026-07-22
CI: Use VecLibFort as shim library for Apple Accelerate/VecLib on macOS.

mmuetzel:
## Pull request purpose

Currently, the CI rules are building ARPACK using the old `f2c` Fortran ABI on macOS. However, all packages of applications written in Fortran that are distributed by Homebrew are using the default GNU Fortran ABI. Testing the package with the ABI expected by most macOS users would probably be better than what is currently done in the CI setup.

Instead of building a version of ARPACK using that old, non-default `f2c` Fortran ABI, use a library that "translates" from the default GNU Fortran ABI to the old `f2c` ABI used by Apple Accelerate/VecLib.

### DIFF
--- a/.github/workflows/jobs.yml
+++ b/.github/workflows/jobs.yml
@@ -176,18 +176,21 @@ jobs:
       - name: Install brew dependencies
         run: |
           brew reinstall gcc # brings gfortran on path
-          brew install cmake open-mpi eigen
+          brew install cmake open-mpi eigen veclibfort
       - name: Run job
         run: |
           mkdir -p build
           cd build
           export FC=mpif90 # Uses gfortran.
-          export FFLAGS="-ff2c -fno-second-underscore"
           export CC=mpicc # Uses clang.
           export CFLAGS="-Qunused-arguments"
           export CXX=mpic++ # Uses clang++.
           export CXXFLAGS="-Qunused-arguments"
-          cmake -DBLA_VENDOR=Apple -DEXAMPLES=ON -DICB=ON -DEIGEN=ON -DMPI=ON ..
+          export VECLIBFORT_PREFIX=$(brew --prefix veclibfort)
+          cmake \
+            -DBLAS_LIBRARIES="${VECLIBFORT_PREFIX}/lib/libvecLibFort.dylib" \
+            -DLAPACK_LIBRARIES="${VECLIBFORT_PREFIX}/lib/libvecLibFort.dylib" \
+            -DEXAMPLES=ON -DICB=ON -DEIGEN=ON -DMPI=ON ..
           make all
           CTEST_OUTPUT_ON_FAILURE=1 make test
   macos_latest_cmake_python:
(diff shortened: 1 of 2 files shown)


.github/workflows/jobs.yml
```diff
@@ -176,18 +176,21 @@ jobs:
...
-          export FFLAGS="-ff2c -fno-second-underscore"
           export CC=mpicc # Uses clang.
           export CFLAGS="-Qunused-arguments"
           export CXX=mpic++ # Uses clang++.
           export CXXFLAGS="-Qunused-arguments"
-          cmake -DBLA_VENDOR=Apple -DEXAMPLES=ON -DICB=ON -DEIGEN=ON -DMPI=ON ..
+          export VECLIBFORT_PREFIX=$(brew --prefix veclibfort)
+          cmake \
+            -DBLAS_LIBRARIES="${VECLIBFORT_PREFIX}/lib/libvecLibFort.dylib" \
+            -DLAPACK_LIBRARIES="${VECLIBFORT_PREFIX}/lib/libvecLibFort.dylib" \
```
fghoussen, line 192: Possible to drop BLAS_LIBRARIES / LAPACK_LIBRARIES using BLA_VENDOR: can cmake handle that?

fghoussen (commented):
Thanks! Can you update  the CHANGELOG?

.github/workflows/jobs.yml
mmuetzel: There is no value of `BLA_VENDOR` that would identify the `vecLibFort` shim library:
https://cmake.org/cmake/help/latest/module/FindBLAS.html#blas-lapack-vendors

Alternatively, it would be possible to use OpenBLAS instead of Apple Accelerate/VecLib in the CI. OpenBLAS is (mostly) written in C implementing an interface that is compatible to the GNU Fortran ABI. So for OpenBLAS, no shim library is necessary.

mmuetzel:
> Can you update the CHANGELOG?

Done.

(Note that is likely introducing a merge conflict with #483.)

.github/workflows/jobs.yml
fghoussen: In the CI, better to use vecLib (OpenBLAS is already tested on other jobs).

> There is no value of BLA_VENDOR

For me Apple means "use vecLib"
image

mmuetzel: > For me Apple means "use vecLib"

Correct. But that is linking to vecLib directly (without the shim library vecLibFort).
I.e., that would cause issues because vecLib uses the old f2c Fortran ABI that should be avoided nowadays.

fghoussen: OK. But does this mean both vecLib (set by BLA_VENDOR?) and vecLibFort must be passed to cmake? Or one out of both?

mmuetzel: If `BLAS_LIBRARIES` and `LAPACK_LIBRARIES` are set, `BLA_VENDOR` doesn't have any effect.
You can set it to anything(?). But that might be confusing imho. It probably fits the situation best to not set it at all.

fghoussen: > fits the situation best to not set it at all.

101% agree. Can you try to get minimal cmake settings? Is ideally setting BLA_VENDOR and not setting any BLAS_LIBRARIES / LAPACK_LIBRARIES possible?

mmuetzel: Like I wrote before: There is no value for `BLA_VENDOR` that would correspond to the `vecLibFort` shim library.
So, this is the minimal set of variables that need to be set.

mmuetzel: The only other alternative is to not use Apple Accelerate/vecLib and use a different implementation of BLAS/LAPACK instead. That could be OpenBLAS (for which CMake implemented an identifier in its `BLA_VENDOR` selection).
But you already wrote that you didn't want that either.

fghoussen:
@mmuetzel: thanks!

Issue #479, opened 2026-03-08
--with-mpi-include= and --with-mpi-lib= are not respected


yurivict:
When we are trying to use openmpi using these config vars:
```
CONFIGURE_ARGS+=        --enable-mpi \
                        --with-mpi=${MPI_PREFIX} \
                        --with-mpi-lib="${MPI_LIBS}" \
                        --with-mpi-include="${MPI_INCLUDES}"
```

these are not respected and it still uses mpich instead of openmpi values.

sylvestre:
patch welcome :)

yurivict:
I am a FreeBSD port maintainer for the math/arpack-ng port and I don't have resources to submit patches.

sylvestre:
ok, not sure anyone will work on it !
sorry

fghoussen:
How MPI_* are defined? Do you have some logs or associated context?
I don't have a FreeBSD: can't help. But if somebody can it will likely need more info

yurivict:
On FreeBSD mpich is installed into /usr/local and openmpi is installed into /usr/local/openmpi.
When we are trying to build against openmpi and fill the above arguments of configure they are ignored.
There are no MPI_* definitions.

I added the workaround in CFLAGS/FFLAGS/FCFLAGS/PATH variables here: https://cgit.freebsd.org/ports/tree/math/arpack-ng/Makefile#n55

mmuetzel:
See #485 for a potential implementation of configure flags that should make it possible to select an implementation of MPI that is installed at a non-default location.

Issue #474, opened 2024-11-20
skip BLAS when not needed


Fabian188:
I need only the lib and no tests and no examples. In that case it is not necessary to search for blas. 

We already have variables TESTS and EXAMPLES in CMakeLists.txt.
My patch simple does not touch anything with BLAS if both are disabled.

I'm not used to github and neither know the process nor the test stuff here.

Probably one of the experience would apply my patch if the content is of interest?

I wrote the patch based on 3.9.1

```
-if (NOT TARGET BLAS::BLAS) # Search only if not already found by upper CMakeLists.txt
+if ((TESTS OR EXAMPLES) AND NOT TARGET BLAS::BLAS) # Search only if required and not already found by upper CMakeLists.txt
     find_package(BLAS REQUIRED)
-
     # BLAS::BLAS target was already created at this point by FindBLAS.cmake if cmake version >= 3.18
     if (NOT TARGET BLAS::BLAS) # Create target "at hand" to ensure compatibility if cmake version < 3.18
         add_library(BLAS::BLAS INTERFACE IMPORTED)
@@ -292,7 +291,7 @@
 
 # Find LAPACK
 
-if (NOT TARGET LAPACK::LAPACK) # Search only if not already found by upper CMakeLists.txt
+if ((TESTS OR EXAMPLES) AND NOT TARGET LAPACK::LAPACK) # Search only if required and not already found by upper CMakeLists.txt
     find_package(LAPACK REQUIRED)
 
     # LAPACK::LAPACK target was already created at this point by FindLAPACK.cmake if cmake version >= 3.18
@@ -304,7 +303,7 @@
 
 # As BLAS/LAPACK does not provide ICB, we may have to deal with symbols the old-fashion-boring-cumbersome-fortran/C way...
 
-if (SYMBOLSUFFIX)
+if ((TESTS OR EXAMPLES) AND SYMBOLSUFFIX)
     if ("${CMAKE_Fortran_COMPILER_ID}" MATCHES "GNU")
       set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp -ffixed-line-length-none")
     elseif ("${CMAKE_Fortran_COMPILER_ID}" MATCHES "Intel")
@@ -388,13 +387,15 @@
 # use -DBUILD_SHARED_LIBS=ON|OFF to control static/shared
[...]

PR #485, opened 2026-07-23
Add support for non-default location of MPI for `configure`

mmuetzel:
## Pull request purpose

Currently, there are no dedicated flags that would point the `configure` script to an implementation of MPI that is installed at a non-default prefix. (See #479.)

Add support for the optional configure flags `--with-mpi-includedir` and `--with-mpi-libdir` to select a non-default installation location of MPI. Additionally, support setting the prefix of the MPI installation with `--enable-mpi=PREFIX`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,7 @@ AM_MAINTAINER_MODE
 
 dnl Checks for standard programs.
 dnl Autoconf was designed with the fundamental fact that compiler (= language support) is NOT an option.
+AC_PROG_F77
 AC_PROG_FC
 AC_PROG_CC
 AC_PROG_CXX
@@ -127,25 +128,42 @@ fi
 
 dnl See if compiling parpack
 AC_ARG_ENABLE([mpi],
-	[AS_HELP_STRING(
-		[--enable-mpi],
-		[build parallel version of arpack with MPI])],
-	[],
-	[AS_VAR_SET([enable_mpi], [no])])
+  [AS_HELP_STRING(
+    [--enable-mpi=PREFIX],
+    [build parallel version of arpack with MPI])],
+  [AS_IF([test x"$enableval" != x"yes"],
+    [AS_VAR_SET([mpi_prefix], [$enableval])])],
+  [AS_VAR_SET([enable_mpi], [no])])
+
 AS_IF([test x"$enable_mpi" != x"no"], [
-	FCFLAGS_SAVE=$FCFLAGS
-	FCFLAGS=""
-	FFLAGS_SAVE=$FFLAGS
-	FFLAGS=""
+  AC_ARG_WITH([mpi-includedir],
+    [AS_HELP_STRING([--with-mpi-includedir=INCDIR],
+      [directory with headers for MPI])],
+    [MPI_CPPFLAGS="-I$withval"],
+    [AS_IF([test x"$mpi_prefix" != x""],
+      [MPI_CPPFLAGS="-I$mpi_prefix/include"])
+    ])
 
-	AC_LANG_PUSH([Fortran 77])
-	AX_MPI([], AC_MSG_ERROR([could not compile a Fortran MPI test program]))
-	AC_SUBST([MPI_Fortran_LIBS], [$MPILIBS])
-	F77=$MPIF77
-	AC_LANG_POP([Fortran 77])
+  AC_ARG_WITH([mpi-libdir],
+    [AS_HELP_STRING([--with-mpi-libdir=LIBDIR],
+      [linker flags for the MPI library])],
+    [MPI_LDFLAGS="-L$withval"],
+    [AS_IF([test x"$mpi_prefix" != x""],
+      [MPI_LDFLAGS="-L$mpi_prefix/lib"])
+    ])
 
-	FCFLAGS=$FCFLAGS_SAVE
-	FFLAGS=$FFLAGS_SAVE
+  FCFLAGS_SAVE=$FCFLAGS
+  FCFLAGS="$FCFLAGS $MPI_CPPFLAGS"
+  FFLAGS_SAVE=$FFLAGS
+  FFLAGS="$FFLAGS $MPI_CPPFLAGS"
+  LDFLAGS_SAVE=$LDFLAGS
+  LDFLAGS="$LDFLAGS $MPI_LDFLAGS"
+
+  AC_LANG_PUSH([Fortran 77])
+  AX_MPI([], AC_MSG_ERROR([could not compile a Fortran MPI test program]))
+  AC_LANG_POP([Fortran 77])
+  MPILIBS="$MPI_LDFLAGS $MPILIBS"
+  F77=$MPIF77
 
   dnl Check if we can use ISO_C_BINDING provided by MPI.
   AC_LANG_PUSH([Fortran 77])
@@ -164,8 +182,7 @@ AS_IF([test x"$enable_mpi" != x"no"], [
                  ],
                  [
                   AC_MSG_RESULT([yes])
-                  FCFLAGS=$FCFLAGS" -DHAVE_MPI_ICB"
-                  FFLAGS=$FFLAGS" -DHAVE_MPI_ICB"
+                  MPI_CPPFLAGS=$MPI_CPPFLAGS" -DHAVE_MPI_ICB"
                   AS_VAR_SET([enable_mpi_icb], [yes])
                  ],
                  [
@@ -175,6 +192,10 @@ AS_IF([test x"$enable_mpi" != x"no"], [
                  ]
                 )
   AC_LANG_POP([Fortran 77])
+
+  FCFLAGS=$FCFLAGS_SAVE
+  FFLAGS=$FFLAGS_SAVE
+  LDFLAGS=$LDFLAGS_SAVE
 
   dnl As MPI can be used with or without ISO_C_BINDING (#ifdef), we need to preprocess code before compiling.
   AC_LANG_PUSH([Fortran 77])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -285,6 +312,12 @@ else
   CPPFLAGS="$CPPFLAGS -Dsnaupd=$snaupdsuff -Dsneupd=$sneupdsuff -Ddnaupd=$dnaupdsuff -Ddneupd=$dneupdsuff"
 fi
 
+dnl Should the MPI preprocessor flags be exported (AC_SUBST) instead
+dnl and only be used in the build rules of PARPACK?
+FFLAGS="$MPI_CPPFLAGS $FFLAGS"
+FCFLAGS="$MPI_CPPFLAGS $FCFLAGS"
+CPPFLAGS="$MPI_CPPFLAGS $CPPFLAGS"
```
mmuetzel, line 319: These flags have previously been prepended directly to `FFLAGS`, `FCFLAGS`, and `CPPFLAGS`.
For the changes here, it made sense to collect the corresponding flags in `MPI_CPPFLAGS` (to avoid replicating the same thing in multiple places).

With that, it would be possible to only use these flags for the library that needs them.

Should this be done?

fghoussen:
Could you add dedicated tests for this? Like apt-get source openmpi, build-from-source mpi and install it in some /opt-or-tmp/mpi directory and build / run arpack test suite? (mpi is a small lib so building it should be OK)

I would understand 101% this work would be above what you intended to do: i.e. adding some kind of ubuntu_latest_autotools_custom_mpi test in the CI.

I would understand 201% you would not consider adding a second test for the same purpose but using cmake: i.e. adding some kind of ubuntu_latest_cmake_custom_mpi test in the CI.

I understand / agree on the need / PR. But, ideally, I'd like this to be _tested_!... With both cmake and autotools as we try to get them both in sync and having the same level of functionality.

I'd like to avoid blind pushing: we had lots of never ending and re-entrant problems with blind pushing... This is why at some point I created   ubuntu_latest_cmake_install / ubuntu_latest_autotools_install and tstCMakeInstall.sh / tstAutotoolsInstall.sh...

mmuetzel:
I agree that a good CI coverage is useful. I can try to look into that.

Would it be ok to split the (currently huge) workflow file into multiple ones (one per platform) first?
That would make it easier to temporarily disable tests for some platforms on my fork while trying to implement that. (I'll likely need a couple iterations until that is setup correctly. And it would save some CI cycles if it would trigger only the tests on Ubuntu.)
GitHub supports disabling (entire) workflows temporarily. But it doesn't give a good option to temporarily disable on a job-in-a-workflow level.

fghoussen:
For now, my advise would be:
1. at your side (on your fork), first remove all CI jobs.
2. then add 2 CI jobs that will run fast
3. when OK, revert  step 3 (on your fork)
4. push force the branch after removing 1 and 3

This is the way I did that when I had to... So you can do what you need without opening a PR (split CI) inside another unrelated one (custom mpi)!

>  I can try to look into that.

Thanks! :)

> split the (currently huge) workflow file into multiple ones first?

Yes but in another PR!... Mixing both would be a mess... I let you decide which way you wanna go

mmuetzel:
> Yes but in another PR!... Mixing both would be a mess... I let you decide which way you wanna go

I opened #489 with a proposed change for that.

I'll mark this PR as draft until that one has settled.

fghoussen:
@mmuetzel: On this PR, could you update the link in the README

image

It seems broken as the yaml of the CI changed.

mmuetzel:
Sorry for the late reply. I'm currently being caught up in other projects.
I'll try to come back to this PR as soon as time permits.

In the meantime, I opened #493 to address your last comment. (Thank you for pointing that out.)

mmuetzel:
I added a CI configuration that builds OpenMPI and installs it at a non-default prefix as you suggested.
I tried to keep the configuration of OpenMPI as close as possible to how Debian/Ubuntu build it. (After all, this isn't about testing OpenMPI configurations. But about whether the new configure flags work.)

Marking as ready for review again.

fghoussen:
Sorry for late reply. I had no time yet to review this PR (and the others): september is busy.

https://github.com/opencollab/arpack-ng/issues/474#issuecomment-5696950778: ideally the same thing should be done for BLAS / LAPACK. I'll try to do that when / if possible...

.github/workflows/build-ubuntu.yml
```diff
@@ -92,23 +92,32 @@ jobs:
...
             install: with
             blas: BLAS
             configure-flags:
-          - tests: with
+          - name: Ubuntu autotools using MKL
+            tests: with
             install: without
             blas: MKL
             configure-flags: --enable-icb --with-blas=mkl_gf_ilp64 --with-lapack=mkl_gf_ilp64
+          - name: Ubuntu autotools with non-default MPI prefix
```
fghoussen, line 115: No time to review. I've just bird-eye the thing.
We fight for years to make autotools/cmake propose the same options/behavior: did you test also cmake with  non-default MPI prefix?

configure.ac
fghoussen: Likely. Old dead thing... Cleaning/simplifying would be nice...

.github/workflows/build-ubuntu.yml
mmuetzel: CMake has its [`FindMPI` module](https://cmake.org/cmake/help/latest/module/FindMPI.html) which handles this. So, there is no need for any changes in ARPACK-ng when using CMake as the build tool.

I don't think we need to add CI tests to check whether CMake works as they document.

Obviously, the flags will have different names. After all, CMake and autotools are quite different build systems.

fghoussen: Yes.. And no?! 

Yes, on a theoretical point of view, I agree: we should not test CMake.
No, on a practical point of view: it's always good to provide examples of "how to use CMake".

I've seen _many_ people raging "find_package doesn't find anything" and blame CMake. CMake rely on the environment to find packages (there is no magic!). So typically, in an admin (that installed MPI or BLAS or LAPACK) and didn't set environment variable _correctly_ in the module-load file then find_package will never find anything and you can _not_ blame CMake! On supercomputers where you are not root, this may become a nightmare...

I believe providing examples of using CMake in these edge-cases (non-default location of some dependency) in the CI is a good thing as people who want to use arpack in these situations will have an example and you could then reply to @Fabian188  here https://github.com/opencollab/arpack-ng/issues/474#issuecomment-5697694313: "did you set at least CMAKE_PREFIX_PATH (or friends) and check your environment is set properly (by module-load or apt or any other machinery that is supposed to have done that for you) like the CI example?"

I understand your point. But, for me, testing also cmake with non-default MPI prefix is valuable.
 
 I let you decide what you think is best.

mmuetzel: In my opinion, adding a CI runner just for that is overkill.
It would just apply to one specific MPI implementation on one specific operating system. Other MPI implementations on other operating systems might need to be handled differently.

Some documentation might be better. But then, it is already documented by CMake. And any instructions here might depend on the CMake version of the user. (When it comes to the FindMPI module specifically, it hasn't changed in a long time. But that doesn't mean that CMake will never change it.)
So, the best might be to just redirect to the official documentation:
https://cmake.org/cmake/help/latest/module/FindMPI.html

And similarly for FindBLAS.

Maybe, having a collection of links to the upstream documentation of the used CMake modules would be best. The respective section of the `README.md` file looks like a good place for that.

fghoussen: > did you test also cmake with non-default MPI prefix?
> I don't think we need to add CI tests to check whether CMake works as they document.
> I let you decide what you think is best.

OK, I can understand both decisions.

mmuetzel: I've opened #496 for that documentation change.

.github/workflows/build-ubuntu.yml
```diff
@@ -92,23 +92,32 @@ jobs:
...
+            tests: with
+            install: without
+            blas: BLAS
+            configure-flags: --enable-mpi=${GITHUB_WORKSPACE}/openmpi --enable-icb --enable-eigen
+            mpi-location: non-default
```
sylvestre, line 120: this only exercises --enable-mpi=PREFIX, the two new --with-mpi-* flags stay untested. please add a case for them too.

.github/workflows/build-ubuntu.yml
```diff
@@ -139,6 +149,44 @@ jobs:
...
+            --disable-silent-rules \
+            --enable-ipv6 \
+            --with-devel-headers \
+            --with-slurm \
+            --with-sge \
+            --without-tm \
+            --prefix=${GITHUB_WORKSPACE}/openmpi
+          echo "::endgroup::"
+          echo "::group::Build OpenMPI"
+          make all
```
sylvestre, line 183: `make all -j$(nproc)` please, building openmpi serially on every run is slow. caching the install would be even better.

configure.ac
```diff
@@ -8,6 +8,7 @@ AM_MAINTAINER_MODE
 
 dnl Checks for standard programs.
 dnl Autoconf was designed with the fundamental fact that compiler (= language support) is NOT an option.
+AC_PROG_F77
```
sylvestre, line 11: why is AC_PROG_F77 needed here? please document the why in the commit message at least.

configure.ac
```diff
@@ -127,25 +128,42 @@ fi
...
+    [AS_HELP_STRING([--with-mpi-includedir=INCDIR],
+      [directory with headers for MPI])],
+    [MPI_CPPFLAGS="-I$withval"],
+    [AS_IF([test x"$mpi_prefix" != x""],
+      [MPI_CPPFLAGS="-I$mpi_prefix/include"])
+    ])
+
+  AC_ARG_WITH([mpi-libdir],
+    [AS_HELP_STRING([--with-mpi-libdir=LIBDIR],
+      [linker flags for the MPI library])],
```
sylvestre, line 149: says "linker flags" but it's a directory, please fix the help string.

configure.ac
```diff
@@ -127,25 +128,42 @@ fi
...
+    [MPI_LDFLAGS="-L$withval"],
+    [AS_IF([test x"$mpi_prefix" != x""],
+      [MPI_LDFLAGS="-L$mpi_prefix/lib"])
```
sylvestre, line 152: hardcoded /lib, what about lib64?

.github/workflows/build-ubuntu.yml
```diff
@@ -139,6 +149,44 @@ jobs:
...
+        if: matrix.mpi-location == 'non-default'
+        run: |
+          echo "::group::Install dependencies of OpenMPI"
+          sudo apt-get install -y \
+            libevent-dev libpmix-dev libucx-dev zlib1g-dev libhwloc-dev \
+            pkgconf libibverbs-dev libfabric-dev libnuma-dev \
+            libpsm-infinipath1-dev libpsm2-dev flex
+          echo "::endgroup::"
+          echo "::group::Download release tarball"
+          curl -o openmpi-5.0.10.tar.bz2 https://download.open-mpi.org/release/open-mpi/v5.0/openmpi-5.0.10.tar.bz2
```
fghoussen: @mmuetzel Can you rather use `apt-get source openmpi-bin`. This avoid relying on curl and hard-coded URL that may change!...

PR #491, opened 2026-08-12
Add CI workflow that cross-builds for several targets on Ubuntu

mmuetzel:
## Pull request purpose

Currently, ARPACK is build for different targets in CI using `qemu` to emulate the compiler. That is costly compared to cross-compiling.
Additionally, the Alpine repository for the PowerPC target got more and more unreliable recently.

Add a new workflow that cross-builds ARPACK on Ubuntu for some of the targets it supports.
Use `qemu` only to run the test suite for targets that require emulation. Select runners that can execute the test suite natively (multilib) if possible.
With that, weed out the workflow matrix for Alpine Linux. While it is probably still a good idea to keep checking that ARPACK works correctly on a MUSL platform, that can be covered with targets that can run natively on the GitHub-hosted runners.
Add a new target for LoongArch64 which is supported by Alpine but not (yet?) by Ubuntu.

This is in preparation for the refactoring of the CI workflows in #489. The main goal is to have all runners pass the CI again before starting that refactoring.

### DIFF
--- a/.github/workflows/build-alpine.yml
+++ b/.github/workflows/build-alpine.yml
@@ -1,4 +1,4 @@
-name: arpack-ng-emu
+name: build-alpine
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


mmuetzel:
With these changes, the CI is all green again.

fghoussen:
@mmuetzel: thanks!

PR #492, opened 2026-09-04
Do not install CMake config files when building with Makefile rules

mmuetzel:
## Pull request purpose

The CMake config files contain the following instruction:
```
 include("${CMAKE_CURRENT_LIST_DIR}/arpackngTargets.cmake")
```

The file `arpackngTargets.cmake` is generated by CMake. It is not generated by the Makefile rules (nor by the `configure` script).

That means that trying to use the CMake config files in downstream CMake projects leads to a fatal error like the following:
```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/arpack-ng/arpackng-config.cmake:35 (include):
  include could not find requested file:
```

CMake does not support catching fatal errors and continuing execution. That means that the configuration for downstream projects bails at that point even if ARPACK-ng is actually correctly installed (apart from the broken CMake config files).

On the other hand, if no CMake config files are installed at all, CMake determines that gracefully in downstream projects. If that should happen, these projects can fall back to alternative ways to locating ARPACK-ng (e.g., using the pkg-config files or looking at default locations).

This reverts #438 to avoid this error.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,7 @@ arpack-ng - next
  * CMake: Fix running CTests on Windows.
  * Use GNU Fortran ABI in CI tests on macOS.
  * Initialize machine dependent constants only once in `PDNAITR`.
+ * [BUG FIX] Do not install (incomplete) CMake config files when building with Makefile rules.
 
 [ Szabolcs Horvát ]
  * fix obsolescent character length warnings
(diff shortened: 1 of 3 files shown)


mmuetzel:
Some of the package repositories seem to have connection problems currently. (Maybe also caused by AI scrapers like in so many other places?)

The proposed changes shouldn't cause any difference in the tests though. And for what they do, they aren't platform dependent. So, if it does the correct thing in one of the autotools jobs (and apparently it does), it will also do it in the other ones.

mmuetzel:
@sylvestre: Does this look ok to you?

PR #496, opened 2026-09-16
Amend documentation for CMake with references to used modules

mmuetzel:
## Pull request purpose

ARPACK-ng uses modules from upstream CMake to find some of its dependencies. 
Especially if these dependencies are not from the same distribution as CMake itself, the user might need to pass additional flags with information for where to find these dependencies. Which flags those are might dependent on the version of CMake, the operating system, or other properties of the build host.

Refer to the upstream documentation of these CMake modules in the `README.md` file. This might help users to find the information that they need to help themselves.


README.md
```diff
@@ -243,7 +243,18 @@ $ make
 $ sudo make install
 ```
 
-✨ Congratulations, you have installed `arpack` lib using cmake (caution: you need `sudo` to install in your system).
+ARPACK-NG uses the following CMake modules to locate dependencies.
+You might need to use additional configuration flags to help these modules find the respective dependencies (especially if they are not from your distribution or if they are installed at a non-default prefix).
+Please, read the documentation of the respective CMake modules for more information about these additional flags:
+* [FindBLAS](https://cmake.org/cmake/help/latest/module/FindBLAS.html)
+* [FindLAPACK](https://cmake.org/cmake/help/latest/module/FindLAPACK.html)
+* [FindMPI](https://cmake.org/cmake/help/latest/module/FindMPI.html) (for PARPACK)
```
sylvestre, line 251: `find_package(Eigen3)` and `PkgConfig` are used too, worth adding?

README.md
```diff
@@ -243,7 +243,18 @@ $ make
...
+* [FindPythonInterp](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html) (for Python3 bindings)
+* [FindPythonLibs](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) (for Python3 bindings)
```
sylvestre, line 253: these three modules were removed in cmake 4.0, so the `latest` links will likely 404 — could you point them at a versioned doc (e.g. /v3.31/)?

fghoussen: We should move to FindPython3?!... Hoping the CI does not break

README.md
```diff
@@ -243,7 +243,35 @@ $ make
...
+* [FindBoost](https://cmake.org/cmake/help/latest/module/FindBoost.html) (for Python3 bindings)
+
+CMake's `find_package` mechanism may use 2 modes:
```
fghoussen: A few more words on CMake mechanics (that everybody may not know...) looks relevant to me but @mmuetzel @sylvestre do not hesitate to revert if you find this overkill

fghoussen: Maybe someday we'll have a conan recipe!... And no more need for this kind of doc :)
image

